PR #1, opened 2020-07-03
First pass at refactoring this to have the same semantics as the maven plugin

sambsnyd:
Not, as you can see, particularly well tested or polished. The plugin can be applied and the tasks registered, and that's basically all I'll guarantee. 

Nothing is done yet with include/exclude, and profiles & activeProfiles probably aren't handled correctly either.


jkschneider:
This gets us so much closer, let's merge and can work off of master.

PR #5, opened 2020-08-28
Discover

SofiaBrittoSchwartz:
Created a new type of task that lists all of the available Recipes and their visitors.


plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
```diff
@@ -170,18 +170,18 @@ protected RefactorPlan plan() {
             );
 
             sourceFiles.addAll(new YamlParser().parse(
-                    getResources().getFiles().stream()
+                    ((String[])getResources().getFiles().stream()
                             .filter(it -> it.isFile() && it.getName().endsWith(".yml") || it.getName().endsWith(".yaml"))
                             .map(File::getAbsolutePath)
-                            .collect(toList())
+                            .collect(toList()).toArray())
```
jkschneider, line 176: We should leave this as a list and provide a second argument which is the root project directory.

PR #21, opened 2021-04-01
Remove rewriteWarn task dependsOn link to check task

aegershman:
In order to be more discreet with usage, unlink rewriteWarn from check by default. rewriteWarn can be wired into dependsOn check through build.gradle config if desired.

May warrant docs update. May need to cross-reference with Maven plugin functionality.

### DIFF
--- a/plugin/src/test/groovy/org/openrewrite/gradle/RewritePluginTest.groovy
+++ b/plugin/src/test/groovy/org/openrewrite/gradle/RewritePluginTest.groovy
@@ -75,30 +75,6 @@ class RewritePluginTest extends RewriteTestBase {
             }
             """.stripIndent()
 
-    def "rewriteWarn task does not have dependsOn link to check task"() {
-        given:
-        projectDir.newFile("settings.gradle")
-        File rewriteYaml = projectDir.newFile("rewrite-config.yml")
-        rewriteYaml.text = rewriteYamlText
-
-        File buildGradleFile = projectDir.newFile("build.gradle")
-        buildGradleFile.text = buildGradleFileText
-
-        when:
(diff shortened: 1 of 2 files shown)


plugin/src/test/groovy/org/openrewrite/gradle/RewritePluginTest.groovy
```diff
@@ -75,8 +75,31 @@ class RewritePluginTest extends RewriteTestBase {
...
+        def checkResult = result.task(":check")
+
+        then:
+        checkResult.outcome == TaskOutcome.UP_TO_DATE
+        // "rewriteWarn" task should not have been called
+        rewriteWarnResult == null
+
+        where:
+        gradleVersion << GRADLE_VERSIONS_UNDER_TEST
+    }
```
sambsnyd, line 100: I don't know if it makes any sense to assert the _absence_ of a dependency relationship... I'd probably just delete this test altogether. The original version of this test asserting the presence of a dependency relationship was also of marginal value

sambsnyd (changes requested):
I think we can delete that test case without losing much of value. Looks good other than that.

plugin/src/test/groovy/org/openrewrite/gradle/RewritePluginTest.groovy
aegershman: Agreed, removed. Although, I do think the original version of the test which asserted the presence of a dependency relationship was a good idea. If for no other reason than for the test to demonstrate the relationship's existence at all in a provable test format.

Ehhh but in any case, yeah, removed 👍  😉

Issue #20, opened 2021-04-01
Update gradle task names, leave behind old names as undocumented aliases for now


sambsnyd:
"warn" and "fix" are legacy names from when we thought quite a bit differently about the intended usage of these things. 

"rewriteFix" should be renamed to "rewriteRun". 
"rewriteWarn" should be renamed to "rewriteDryRun". The existing behavior only prints which files would be changed by which visitors, which isn't very useful. It should instead print what the actual changes would be, similar to what `git diff` would show you after running rewriteRun/rewriteFix

aegershman:
moved changing rewriteDryRun's behavior to a separate issue https://github.com/openrewrite/rewrite-gradle-plugin/issues/27

okundzich:
Why are we leaving undocumented aliases for now? And at what point do we want to remove them?

Also, rewriteDryRun could produce a very long diff if it produces full output rather than a list of files changed. git diff paginates it, but build plugins do not. Maybe leaving existing behaviour is better.

aegershman:
I will change the title for this and edit the original issue body-- we aren't leaving undocumented aliases any longer, the names are fully changed.

As for changes in rewriteDryRun output, I went ahead and created a separate issue in https://github.com/openrewrite/rewrite-gradle-plugin/issues/27

PR #25, opened 2021-04-01
rename tasks rewriteWarn to rewriteDryRun and rewriteFix to rewriteRun

aegershman:
closes https://github.com/openrewrite/rewrite-gradle-plugin/issues/20

should likely be followed by https://github.com/openrewrite/rewrite-maven-plugin/issues/110

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -26,8 +26,8 @@ import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.TaskContainer;
 
 /**
- * Adds the RewriteExtension to the current project and registers tasks per-sourceSet that implement rewrite fixing and
- * warning. Only needs to be applied to projects with java sources. No point in applying this to any project that does
+ * Adds the RewriteExtension to the current project and registers tasks per-sourceSet.
+ * Only needs to be applied to projects with java sources. No point in applying this to any project that does
  * not have java sources of its own, such as the root project in a multi-project builds.
  */
 public class RewritePlugin implements Plugin<Project> {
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -59,15 +59,14 @@ public void apply(Project project) {
...
-        // Fix is meant to be invoked manually and so is not made a dependency of any existing task
         Task rewriteFixAll = tasks.create("rewriteFix",
                 taskClosure(task -> {
                     task.setGroup("rewrite");
                     task.setDescription("Apply the active refactoring recipes to all sources");
                 })
         );
 
-        Task rewriteWarnAll = tasks.create("rewriteWarn", taskClosure(task -> {
+        Task rewriteDryRunAll = tasks.create("rewriteDryRun", taskClosure(task -> {
```
sambsnyd, line 69: Let's keep the old "warn" name working for now, just in case someone uses it. Create another task named rewriteWarn that depends on `rewriteDryRun`. We'll leave the aliases in for a few releases

aegershman: Just to clarify: are we just aliasing `rewriteWarn`, or do we also need to alias `rewriteWarn{SourceSet}`, e.g. users running `rewriteWarnTest`, `rewriteWarnMain`, etc.?

sambsnyd: Probably just `rewriteWarn`. I think it unlikely that anyone will have bothered with the more granular per-sourceSet tasks

aegershman: Just for good measure I included the specific source sets, too.

aegershman:
I can include aliasing rewriteFix in this same PR as well, just wanted to get a PR to work out the general details

aegershman:
(going to close this PR outright and will open a separate one renaming both rewriteWarn and rewriteFix in the same PR and without aliasing)

PR #41, opened 2021-05-01
Isolate classloader for Rewrite dependencies

### LINKED ISSUE #40: Isolate classpath/classloader for rewrite from other gradle plugins
All gradle plugins included in the same build share the same classpath at runtime.
This can be problematic when plugins used in the same build depend on different versions of the same dependencies. We've persistently had issues with our dependency on jackson conflicting with other gradle plugins that also depend on jackson.

To resolve this, we should remove our compile-time dependencies on rewrite and instead construct a detached Gradle dependency configuration at runtime, resolve the rewrite dependencies there, and construct our own isolated classloader through which we load & use rewrite at runtime.

sambsnyd:
Now all rewrite interaction is isolated to its own classloader behind a reflective facade.
Creating the facade was a bit tedious, couldn't help feel like there must be a better way to do this sort of thing. Open to suggestion about better ways of going about this. 

Closes #40


tkvangorder (approved):
Looks reasonable to me, nice job sorting this out.

aegershman (approved):

PR #48, opened 2021-05-05
set default extension value of rewrite to 7.3.0

aegershman:


sambsnyd (approved):

plugin/src/main/java/org/openrewrite/gradle/RewriteExtension.java
```diff
@@ -33,7 +33,7 @@
     private final Project project;
     private File configFile;
     private String metricsUri = magicalMetricsLogString;
-    private String rewriteVersion = "7.2.2";
+    private String rewriteVersion = "7.3.0";
```
aegershman, line 36: @sambsnyd just to double-check, this is technically all that should be necessary to "bump the version" of `rewrite` core used to 7.3.0?

aegershman: going to guess that the exact same time review approval means "yes" ; )

PR #51, opened 2021-05-05
Prevent DryRun from caching .patch files

aegershman:
do not merge, putting here for discussion + thought related to whether it makes sense for subsequent calls to dryRun to output 'up-to-date' even if dryrun technically _would_ make changes.


aegershman:
thinking on this over in https://github.com/openrewrite/rewrite-gradle-plugin/pull/52

Issue #42, opened 2021-05-01
Expose option to fail build if `rewriteDryRun` fails


jbduncan:
At the moment, if one runs `gradle rewriteDryRun`, then it reports any code that hasn't been refactored.

However, I'd love for the build to fail if `rewriteDryRun` reports any such code, so that I know about it at CI time.

It seems to me that a good way forward would be to expose a flag in the extension that, when enabled, makes the build fail if a dry run reports any problems. It would be `false` by default to maintain backwards compatibility. So something like:

```kotlin
rewrite {
  activeRecipe("com.yourorg.TheRecipe")

  failOnBadDryRun = true
}
```

But I'm open to other ideas!

What do you think, Rewrite team?

aegershman:
👍   right on, thank you for logging this. We'll circle back to you on this.

(for issue-management purposes, linking the equivalent `rewrite-maven-plugin` logged issue: https://github.com/openrewrite/rewrite-maven-plugin/issues/139)

jbduncan:
To add to my first comment, I found yesterday that one can apparently workaround this by configuring CI to check if any of the task's warnings are logged: https://docs.openrewrite.org/reference/gradle-plugin-configuration#the-dryrun-task

But, of course, the advantage of having this issue solved is it would allow `rewriteDryRun` to fail even outside of CI. :)

aegershman:
Hey @jbduncan with https://github.com/openrewrite/rewrite-gradle-plugin/releases/tag/v4.2.1 you can wire up this configuration to have `rewriteDryRun` throw an exception/non-zero exit code on any detected changes:

```kt
rewrite {
    failOnDryRunResults = true
}
```

I'll add it to the reference docs a bit later. I'm going to close out this issue, but feel free to comment on this or open a different issue otherwise 👍

jbduncan:
@aegershman Wow, thank you for such a quick turnaround!

I can confirm that version 4.2.2 has a `failOnDryRunResults` flag and, when I enable it, then change the ordering of a couple of modifiers and apply [`org.openrewrite.java.cleanup.ModifierOrder`](org/openrewrite/java/cleanup/ModifierOrder.java), then it causes my build to fail.

:tada:

PR #52, opened 2021-05-05
Configurable failOnDryRunResults

aegershman:
Addresses https://github.com/openrewrite/rewrite-gradle-plugin/issues/42 and https://github.com/openrewrite/rewrite-maven-plugin/issues/139

Configurable flag for determining whether to fail on dry run results. Currently, default is not to fail. This allows a user to configure having dryRun outputting changes result in an exception and thus non-zero exit code.

```kt
rewrite {
    failOnDryRunResults = true
}
```

Example in the case of dryRun producing a result set:

```log
Running recipe(s)...
These recipes would make results to src/main/java/io/slugstack/oss/unspecifiedgradleproject/UnspecifiedGradleProjectApplication.java:
    org.openrewrite.java.format.AutoFormat
Report available:
    /Users/aegershman/projects/slugstack/slugstack-unspecified-gradle-project/build/reports/rewrite/main.patch
Run 'mvn rewrite:run' to apply the recipes.

> Task :rewriteDryRunMain FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':rewriteDryRunMain'.
> Result changes detected. Please see result file for more information.
```

or the default (as in, `false`), which is the status quo:

```log
Running recipe(s)...
These recipes would make results to src/test/java/io/slugstack/oss/unspecifiedgradleproject/UnspecifiedGradleProjectApplicationTests.java:
    org.openrewrite.java.format.AutoFormat
Report available:
    /Users/aegershman/projects/slugstack/slugstack-unspecified-gradle-project/build/reports/rewrite/test.patch
Run 'mvn rewrite:run' to apply the recipes.

BUILD SUCCESSFUL in 8s
4 actionable tasks: 2 executed, 2 up-to-date
```

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewriteDryRunTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewriteDryRunTask.java
@@ -19,6 +19,7 @@ import org.gradle.api.GradleException;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
+import org.gradle.api.specs.Specs;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskAction;
@@ -44,7 +45,7 @@ public class RewriteDryRunTask extends AbstractRewriteTask {
         super(configuration, sourceSet, extension);
         setGroup("rewrite");
         setDescription("Dry run the active refactoring recipes to sources within the " + sourceSet.getName() + " SourceSet. No results will be made.");
-        getOutputs().upToDateWhen(task -> !extension.getFailOnDryRunResults());
+        getOutputs().upToDateWhen(Specs.SATISFIES_NONE);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewriteDryRunTask.java
```diff
@@ -43,6 +44,7 @@ public RewriteDryRunTask(Configuration configuration, SourceSet sourceSet, Rewri
         super(configuration, sourceSet, extension);
         setGroup("rewrite");
         setDescription("Dry run the active refactoring recipes to sources within the " + sourceSet.getName() + " SourceSet. No results will be made.");
+        getOutputs().upToDateWhen(task -> !extension.getFailOnDryRunResults());
```
aegershman, line 47: Maybe better to just have this be `getOutputs().upToDateWhen(Specs. SATISFIES_NONE)` to prevent the output of `dryRun` from ever being cached. That might be simpler, less surprising, and keep functionality in-line with what the maven-plugin will be 🤔

aegershman: Actually, going to invert it. Let's just not cache the result. It's overall less surprising this way.

PR #63, opened 2021-08-03
Adding Java provenance info to gradle-based builds

### LINKED ISSUE #62: Add JavaProvenance information to ASTs parsed by the gradle plugin

tkvangorder:
Fixes #62

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
@@ -21,6 +21,7 @@ import org.gradle.api.logging.Logger;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.SourceSet;
+import org.gradle.util.GradleVersion;
 import org.openrewrite.gradle.RewriteReflectiveFacade.*;
 
 import java.io.File;
@@ -30,6 +31,7 @@ import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
+import java.util.function.UnaryOperator;
 
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
@@ -38,11 +40,11 @@ import static java.util.stream.Collectors.toMap;
 public abstract class AbstractRewriteTask extends DefaultTask implements RewriteTask {
 
     private final Configuration configuration;
-    private final Collection<SourceSet> sourceSets;
+    private final Map<SourceSet, RewriteJavaMetadata> sourceSets;
     protected final RewriteExtension extension;
     private final RewriteReflectiveFacade rewrite;
 
-    public AbstractRewriteTask(Configuration configuration, Collection<SourceSet> sourceSets, RewriteExtension extension) {
+    public AbstractRewriteTask(Configuration configuration, Map<SourceSet, RewriteJavaMetadata> sourceSets, RewriteExtension extension) {
         this.configuration = configuration;
         this.extension = extension;
         this.sourceSets = sourceSets;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -80,9 +80,18 @@ public void apply(Project rootProject) {
                 if(!(plugin instanceof JavaBasePlugin)) {
                     return;
                 }
-                JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention.class);
-                javaConvention.getSourceSets().all(sourceSet -> {
-                    sourceSets.add(sourceSet);
+                //Collect Java metadata for each project (used for Java Provenance)
+                JavaPluginExtension javaPluginExtension = project.getExtensions().getByType(JavaPluginExtension.class);
```
sambsnyd, line 84: `JavaPluginExtension` was added in Gradle 4.10, which is newer than the oldest version we currently support (4.7), and we were talking about going back to support even earlier versions of Gradle. Ultimately to support Gradle versions new and old we'll need to use reflection to load one or the other at runtime. For now stick with `JavaPluginConvention` if you can.

plugin/src/main/java/org/openrewrite/gradle/RewriteReflectiveFacade.java
```diff
@@ -193,6 +193,160 @@ public String print() {
...
+        if (sourceFileGetMarkers == null) {
+            try {
+                Class<?> sourceFileClass = getClassLoader().loadClass("org.openrewrite.SourceFile");
+                sourceFileGetMarkers = sourceFileClass.getMethod("getMarkers");
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return sourceFileGetMarkers;
+    }
```
sambsnyd, line 220: Why isn't this `RewriteReflectiveFacade.SourceFile.getMarkers()`? Same question re: `sourceFileWithMarkersMethod()` and `markersAddIfAbsentMethod()`.

This is inconsistent with how the rest of the facade is laid out and I don't see why that should be.

plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
```diff
@@ -176,17 +178,30 @@ protected ResultsContainer listResults() {
...
                     .collect(toList());
 
+            String javaRuntimeVersion = System.getProperty("java.runtime.version");
+            RewriteJavaMetadata javaMetadata =  sourceSets.get(sourceSet);
+            JavaProvenance javaProvenance = rewrite.javaProvenanceBuilder()
+                    .projectName(getProject().getName())
+                    .buildToolVersion(GradleVersion.current().getVersion())
+                    .sourceSetName(sourceSet.getName())
+                    .vmRuntimeVersion(javaRuntimeVersion)
+                    .vmRuntimeVersion(System.getProperty("java.vm.vendor"))
```
sambsnyd, line 188: This looks accidental. You probably didn't mean to set the same property twice with different values

sambsnyd (changes requested):
Looks good overall 👍

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
tkvangorder: makes sense

plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
tkvangorder: yep, should be vmVendor, good catch

plugin/src/main/java/org/openrewrite/gradle/RewriteReflectiveFacade.java
tkvangorder: I honestly was trying to avoid creating:

A facade for Markers, and a polymorphic facade for Marker and then having JavaProvenance extend that. 

When in fact we know that there is a very limited use case for adding the JavaProvenance

PR #73, opened 2021-09-01
Clean up thread resources on task completion

tkvangorder:


tkvangorder:
This was addressed in #74

PR #74, opened 2021-09-02
Add ability to cache ASTs in-memory when using the Gradle daemon

tkvangorder:


tkvangorder:
This PR also addresses some additional resource cleanup after Rewrite's tasks have been executed.

PR #124, opened 2022-07-26
Simplify build and further isolate rewrite proper

shanman190:


shanman190:
NOTE: Will want to delete the files from your local `src/main/resources` folder that represent the properties files. Shoving these into `src/main/resources` has implications on task cacheability, much of which can be avoided with the solution provided here.

PR #156, opened 2023-01-24
Added KotlinParser to Resourceparser.

traceyyoshima:


sambsnyd:
As you can tell this wont set up classpaths and dependencies and such. So any ASTs coming out of this will be lacking type attribution. Is that valuable for you? In a bit I can see about setting up the classpath appropriately

traceyyoshima:
> As you can tell this wont set up classpaths and dependencies and such. So any ASTs coming out of this will be lacking type attribution. Is that valuable for you? In a bit I can see about setting up the classpath appropriately

Closing PR. The base classpath is already set up in the parser. I can test locally when I find sample projects or create a small project to test on.

PR #162, opened 2023-02-06
#412 - Added OS detection to Provenance

reka18:


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -884,4 +891,13 @@ private void logRecipe(RecipeDescriptor rd, String prefix) {
             logRecipe(rChild, prefix + "    ");
         }
     }
+
+    private OperatingSystem detectOs() {
+        String osStr = SystemUtils.OS_NAME.toLowerCase();
+        if (osStr.contains("windows")) {
```
sjungling, line 897:
```suggestion
        if (OperatingSystem.current().isWindows()) {
```

sjungling (commented):
can use a helper from that class?

sjungling (approved):

PR #174, opened 2023-02-24
Added org.jetbrains.kotlin.multiplatform to accepted plugins in the DefaultProjectParser.

### LINKED ISSUE #172: Plugin does not generate Kotlin sources on Kotlin project
The `DefaultProjectParser` will parse sources with the KotlinParser if the `org.jetbrains.kotlin.jvm` exists in the project.
`arrow-kt` is a Kotlin project primarily uses the `kotlin.multiplatform` plugin.

An analysis of sources with the `.kt` using data tables implies that most Kotlin projects do not use the `jvm` project and use a different plugin.

This issue will track supporting the Gradle plugin to add sources with the `kotlin.multiplatform` plugin.

https://github.com/arrow-kt/arrow/blob/main/build.gradle.kts

traceyyoshima:
Changes:

- subprojects with `"org.jetbrains.kotlin.multiplatform"` plugin will generate Kotlin source files.

fixes #172

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -682,7 +682,7 @@ public class DefaultProjectParser implements GradleProjectParser {
                     }
                 }
 
-                if(subproject.getPlugins().hasPlugin("org.jetbrains.kotlin.jvm")) {
+                if (subproject.getPlugins().hasPlugin("org.jetbrains.kotlin.jvm") || subproject.getPlugins().hasPlugin("org.jetbrains.kotlin.multiplatform")) {
                     List<Path> kotlinPaths = sourceSet.getAllSource().getFiles().stream()
                             .filter(it -> it.isFile() && it.getName().endsWith(".kt"))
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -682,7 +682,7 @@ public List<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Execu
                     }
                 }
 
-                if(subproject.getPlugins().hasPlugin("org.jetbrains.kotlin.jvm")) {
```
sambsnyd, line 685: I would be very surprised if this leads to a single new kotlin source being parsed. 

If you scroll up, you'll see we're querying the Java plugin's concept of `SourceSet`, which the kotlin JVM plugin builds on top of, to extract information. This is inside a loop over the Java `SourceSet`s. 

I don't believe the kotlin multiplatform, native, or js plugins are built on top of the java plugin at all.
They likely have an analogous concept with a distinct implementation from java source sets.

traceyyoshima:
Closing PR, will reevaluate with a different approach

PR #185, opened 2023-03-25
refactor: SLF4J best practices

timtebeek:
Spurred on by https://github.com/openrewrite/rewrite/issues/3032 I thought to apply the best practice recipes where they made notable changes. Running these now also ensures we're in a better state when we add pull request checks on no-outstanding-changes.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/ResourceParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/ResourceParser.java
@@ -171,8 +171,7 @@ public class ResourceParser {
                 if (!attrs.isOther() && !attrs.isSymbolicLink() &&
                         !alreadyParsed.contains(file) && !isExcluded(file)) {
                     if (isOverSizeThreshold(attrs.size())) {
-                        logger.info("Parsing as quark " + file + " as its size + " + attrs.size() / (1024L * 1024L) +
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/ResourceParser.java
```diff
@@ -171,8 +171,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
                 if (!attrs.isOther() && !attrs.isSymbolicLink() &&
                         !alreadyParsed.contains(file) && !isExcluded(file)) {
                     if (isOverSizeThreshold(attrs.size())) {
-                        logger.info("Parsing as quark " + file + " as its size + " + attrs.size() / (1024L * 1024L) +
-                                "Mb exceeds size threshold " + sizeThresholdMb + "Mb");
+                        logger.info("Parsing as quark {} as its size + {}Mb exceeds size threshold {}Mb", file, attrs.size() / (1024L * 1024L), sizeThresholdMb);
```
timtebeek, line 174:
```suggestion
                        logger.info("Parsing as quark {} as its size {}Mb exceeds size threshold {}Mb", file, attrs.size() / (1024L * 1024L), sizeThresholdMb);
```

PR #191, opened 2023-04-11
Correctly and completely manage source set assignment to all compilation units

shanman190:
Relates to https://github.com/openrewrite/rewrite/pull/3105 and https://github.com/openrewrite/rewrite-kotlin/pull/95.

Fixes https://github.com/openrewrite/rewrite/issues/3053.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -57,14 +57,11 @@ import org.openrewrite.java.style.Autodetect;
 import org.openrewrite.java.style.CheckstyleConfigLoader;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.kotlin.KotlinParser;
 import org.openrewrite.kotlin.tree.K;
-import org.openrewrite.marker.BuildTool;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -705,17 +701,20 @@ public List<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Execu
...
+                    List<JavaType.FullyQualified> classpath = sourceSetProvenance.getClasspath();
+                    for (J.CompilationUnit cu : cus) {
+                        for (JavaType type : cu.getTypesInUse().getTypesInUse()) {
+                            if (type instanceof JavaType.FullyQualified) {
+                                classpath.add((JavaType.FullyQualified) type);
+                            }
+                        }
+                    }
+                    JavaSourceSet javaSourceSetProvenance = sourceSetProvenance.withClasspath(classpath);
+                    sourceFiles.addAll(map(autodetectStyle(cus), addProvenance(projectProvenance, javaSourceSetProvenance)));
```
sambsnyd, line 713: All java/kotlin/groovy types are visible to eachother within the source set (assuming you're using java + groovy OR java + kotlin, all three mixed together makes for difficult build configurations in my experience). 

What if we don't put any provenance markers on until all three sources of types have been parsed, with their types added to a single set (to avoid duplicate entries), then all sources are given the same instance of the same `JavaSourceSet` all at once? Feels like that would be most efficient / de-duplicated.

Issue #196, opened 2023-04-30
Documentation 404


koppor:
The lower part of README.md shows:

![image](https://user-images.githubusercontent.com/1366654/235372626-349b7a37-ba72-4abe-85ed-15dbbc006860.png)

The second link is 404: https://docs.openrewrite.org/reference/gradle-plugin-configuration

The homepage does not find "gradle plugin reference":

![image](https://user-images.githubusercontent.com/1366654/235372643-d9a4117d-b39d-48ee-82d0-9dc4c19908b9.png)

(Link: https://docs.openrewrite.org/running-recipes/getting-started?q=gradle+plugin+reference)

Source of the doc still exists: https://github.com/openrewrite/rewrite-docs/blob/master/reference/gradle-plugin-configuration.md

timtebeek:
@mike-solomon Any idea why GitBook doesn't deploy that Gradle plugin reference?
The link on the reference page of the documentation does not resolve to docs, but to GitHub:
https://docs.openrewrite.org/reference links to https://github.com/openrewrite/rewrite-docs/blob/master/reference/gradle-plugin-configuration.md
If we manage to get the deployment of the reference page fixed, then we can also close this issue.
If not we can look at alternatives, such as again using GitHub Pages for deployment of generated plugin documentation.

PR #197, opened 2023-04-30
Fix link

koppor:
The lower part of README.md shows:

image

Both links are 404. This PR fixes the first one.

For the second one, I could not find anything (and therefore opened an isse https://github.com/openrewrite/rewrite-gradle-plugin/issues/196).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,5 +60,5 @@ plugins {
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


timtebeek (approved):

timtebeek:
Thanks a lot @koppor ! The other failure is an odd one, we'll tackle that separately.

PR #200, opened 2023-05-11
OpenRewrite 8.0 upgrade

knutwannheden:


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -595,27 +591,26 @@ protected Environment environment() {
...
-    public List<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, ExecutionContext ctx) {
+    public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, ExecutionContext ctx) {
         Collection<PathMatcher> exclusions = extension.getExclusions().stream()
                 .map(pattern -> subproject.getProjectDir().toPath().getFileSystem().getPathMatcher("glob:" + pattern))
                 .collect(toList());
-        if(isExcluded(exclusions, subproject.getProjectDir().toPath())) {
+        if (isExcluded(exclusions, subproject.getProjectDir().toPath())) {
             logger.lifecycle("Skipping project {} because it is excluded", subproject.getName());
-            return Collections.emptyList();
+            return Stream.of();
```
shanman190, line 613: Would it be better to use `Stream.empty()` for these cases?

PR #204, opened 2023-06-09
Remove global shared `JavaTypeCache`

knutwannheden:
The JVM-based parsers should not all share the same `JavaTypeCache`, especially not in multi-module builds or across source set boundaries.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -87,7 +87,6 @@ import static org.openrewrite.Tree.randomId;
 
 @SuppressWarnings("unused")
 public class DefaultProjectParser implements GradleProjectParser {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/openrewrite/gradle/DelegatingProjectParser.java
```diff
@@ -125,6 +125,8 @@ protected URL jarContainingResource(String resourcePath) {
                     resourcePath = resourcePath.substring(0, indexOfBang);
                 }
                 return new URI(resourcePath).toURL();
+            } else if (resourcePath.startsWith("file:")) {
+                return new URI(resourcePath.substring(0, resourcePath.lastIndexOf("/main/") + 6)).toURL();
```
jkschneider, line 129: What is this?

knutwannheden: That is an unrelated change I should commit separately. Without this I cannot debug the tests using the debugger...

PR #206, opened 2023-06-13
Correct grammar mistake in dry run reporting output.

jeremyg484:
I noticed and corrected a grammatical error in the `rewriteDryRun` task output.


sambsnyd:
Thanks @jeremyg484 !

PR #213, opened 2023-07-07
Fixes invalid exclusion paths

rpau:
## What's changed?
The exclusion paths were calculated from the parent project path instead of the base dir. This has caused duplicated source files in the Moderne Gradle plugin.

## What's your motivation?
Duplicated source files in the LSTs we generate.

- I required to add the rewrite dependencies as a testRuntime dependencies to properly add a test that was failing before and now succeeds. 

@knutwannheden 
@sambsnyd 
@jkschneider


plugin/src/test/kotlin/org/openrewrite/gradle/isolated/ResourceParserTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.openrewrite.gradle.DefaultRewriteExtension
+import org.openrewrite.java.internal.JavaTypeCache
+import kotlin.io.path.Path
+
+class ResourceParserTest {
+
+    @Test
+    fun `resource parser is excluding subprojects directories using the base dir`(
+    ) {
+        val project = Mockito.mock(Project::class.java)
```
shanman190, line 31: Gradle's `ProjectBuilder` is usually how I handle cases such as this.

rpau: Thanks @shanman190 . I have added a new commit that uses `ProjectBuilder` instead of `Mockito`

rpau:
To have a green build, we need to merge https://github.com/openrewrite/rewrite-gradle-plugin/pull/214 first because the main branch is broken

PR #214, opened 2023-07-07
Fixes the broken build for RewriteRunTest#deleteEmptyDirectory

rpau:
## What's changed?

I am adding the groovy plugin to the specific Gradle configuration that is created in the test that is failing. It is the main difference between the failed test and the others that succeed. 
Otherwise, the main branch is broken: https://github.com/openrewrite/rewrite-gradle-plugin/actions/runs/5441989672/jobs/9998886947

This seems related to the python project and the latest changes to introduce the new guava dependency
https://github.com/openrewrite/rewrite-python/commits/main. 

```
Could not resolve com.google.guava:guava:32.1.1-jre.

         Required by:

             project : > org.openrewrite:rewrite-python:1.1.0-SNAPSHOT:20230706.182507-37
```

I do not know if this is a required/expected change or not. 

@knutwannheden 
@timtebeek 

Revert the change in python, but the previous one was a version that contains vulnerabilities.

### DIFF
--- a/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
+++ b/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
@@ -752,6 +752,7 @@ class RewriteRunTest : RewritePluginTest {
             propertiesFile("foo/foo.properties", "foo = bar")
             buildGradle("""
                 plugins {
+                    id("groovy")
                     id("org.openrewrite.rewrite")
                 }
                 


plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
```diff
@@ -752,6 +752,7 @@ class RewriteRunTest : RewritePluginTest {
             propertiesFile("foo/foo.properties", "foo = bar")
             buildGradle("""
                 plugins {
+                    id("groovy")
```
rpau, line 755: I am worry that this becomes a requirement just because of guava.

knutwannheden: That would be strange. I don't see a connection here.

shanman190: So the reason for this is that guava is publishing two variants: android and jre. As part of the Gradle module metadata it shows both variants here.
https://repo.maven.apache.org/maven2/com/google/guava/guava/32.1.1-jre/guava-32.1.1-jre.module

Now the issue, is that with only the rewrite plugin applied to that test sample project, Gradle isn't able to select a single component that satisfies the required capabilities (`com.google.guava:gauva:32.1.1-jre`, `com.google.collections:google-collections:32.1.1-jre`, and `com.google.guava:listenablefuture:1.0`). As such it throws the error observed in the console log.

When you apply the `groovy` plugin, it transitively applies several other plugins, but the key one in particular that resolves the error is the `JavaBasePlugin`/`java-base`. This plugin in particular sets the `org.gradle.jvm.environment` attribute on the configurations which then enables Gradle to determine a _single_ matching dependency (android isn't allowed for the `standard-jvm` value).

You can see that selection being configured here:
https://github.com/gradle/gradle/blob/0d56d9a09105ca260b85b748703ca2bcfce536ed/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L282
https://github.com/gradle/gradle/blob/0d56d9a09105ca260b85b748703ca2bcfce536ed/subprojects/plugins/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmPluginServices.java#L85
https://github.com/gradle/gradle/blob/0d56d9a09105ca260b85b748703ca2bcfce536ed/subprojects/plugins/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmEcosystemAttributesDetails.java#L112

knutwannheden: Ouch. Thanks for clarifying!

shanman190: You're welcome.

If you're super curious, the plugin structure goes like this:

`groovy` -applies-> `java` -applies-> `java-base`

So with that noted, you'll notice that every other test except this one already was applying the `java` plugin which would have added that same attribute fix for each of those tests. This is why we only observed the error on this single test case.

PR #220, opened 2023-08-09
Use OmniParser

jkschneider:


plugin/build.gradle.kts
```diff
@@ -106,13 +106,17 @@ dependencies {
...
+        isTransitive = false
+    }
     "rewriteDependencies"("org.openrewrite.gradle.tooling:model:$latest")
     "rewriteDependencies"("org.openrewrite:rewrite-maven")
     // Newer versions of checkstyle are compiled with a newer version of Java than is supported with gradle 4.x
     "rewriteDependencies"("com.puppycrawl.tools:checkstyle:9.3")
     "rewriteDependencies"("com.fasterxml.jackson.module:jackson-module-kotlin:latest.release")
 
     implementation(platform("org.openrewrite:rewrite-bom:$latest"))
+    implementation("org.openrewrite.recipe:rewrite-all:$latest")
```
shanman190, line 119: I think the use case for this is covered by the `compileOnly` and would otherwise be leaking the OpenRewrite libraries back onto the plugin classpath.

plugin/build.gradle.kts
```diff
@@ -126,6 +130,9 @@ dependencies {
     compileOnly("org.openrewrite:rewrite-protobuf")
     compileOnly("org.openrewrite:rewrite-xml")
     compileOnly("org.openrewrite:rewrite-yaml")
+    compileOnly("org.openrewrite.recipe:rewrite-all:$latest") {
+        isTransitive = false
```
shanman190, line 134: Is disabling transitives actually necessary for some reason?

sambsnyd: Not necessary. Imagine the idea was to just bring in `OmniParser` without all the other recipe modules

plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
```diff
@@ -73,6 +73,7 @@ public Set<File> getResolvedDependencies() {
                     deps.create("org.openrewrite:rewrite-protobuf:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-xml:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-yaml:" + rewriteVersion),
+                    deps.create("org.openrewrite.recipe:rewrite-all:" + extension.getRewriteAllVersion()),
```
shanman190, line 76: @sambsnyd, speaking of `rewrite-all` depending on all recipe libraries... This by default now kinda means that end users now have all recipes available. Are we intentionally wanting for that to be the case or should we disable transitives here, specifically?

Issue #233, opened 2023-10-05
Specifying a rewriteVersion results in rewriteRun failing to resolve dependencies


MagicalAsh:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.6.1
- Maven/Gradle plugin v6.3.11

## How are you running OpenRewrite?

I am using the Gradle plugin, and my project is a single module project.
```groovy
plugins {
    id 'java'
    id 'org.openrewrite.rewrite'
}

rewrite {
    rewriteVersion = "8.6.1"
}
```

## What is the smallest, simplest way to reproduce the problem?

Specifying the rewriteVersion property to be anything other than null causes the issue.
```groovy
rewrite {
    rewriteVersion = "8.6.1"
}
```

## What did you expect to see?

A successful run of OpenRewrite.

## What did you see instead?

An unsuccessful run.

## What is the full stack trace of any errors you encountered?

```
Execution failed for task ':rewriteResolveDependencies'.
> Could not resolve all files for configuration ':detachedConfiguration1'.
   > Could not find org.openrewrite:rewrite-polyglot:8.6.1.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/openrewrite/rewrite-polyglot/8.6.1/rewrite-polyglot-8.6.1.pom
     Required by:
         project :
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

I'd like to, but the policy regarding Open source contributions is a bit long and this should be a very quick change. 

I can tell you that the problem is [here](https://github.com/openrewrite/rewrite-gradle-plugin/blob/5e7be11f8dbf48ff5289daffbcca2f64b008e5b4/plugin/src/main/java/org/openrewrite/gradle/DefaultRewriteExtension.java#L221) as the rewrite-polyglot version is using the same field as the rest of the rewrite bom, despite rewrite-polyglot being versioned independently.

PR #235, opened 2023-10-10
fix: use separate rewrite-polyglot version (#233)

MagicalAsh:
This commit fixes issue #233, where specifying a
rewriteVersion inside the rewrite configuration block would result in rewrite-polyglot failing to resolve to a proper version.

## What's changed?

The rewrite-polyglot version can be configured independently from the rewrite version.

## What's your motivation?

See #233.

### Checklist
- [x] I've added the license header to any new files through `./gradlew licenseFormat`

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/DefaultRewriteExtension.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/DefaultRewriteExtension.java
@@ -217,12 +217,13 @@ public class DefaultRewriteExtension implements RewriteExtension {
         return rewriteVersion;
     }
 
+    private String rewritePolyglotVersion;
     @Override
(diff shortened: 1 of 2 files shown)


timtebeek:
This seems to have been a bug of omission indeed, especially looking at how we have `rewriteGradleModelVersion` and `rewriteKotlinVersion` just below. Thanks for stepping up and getting that resolved!

timtebeek (approved):

MagicalAsh:
Awesome, thanks! Do you have a timeline on when this will be available in a release version?

timtebeek:
I'd expect a new release somewhere this week or next. There was one change we were waiting for, but I believe that hurdle had been cleared. You can already try the snapshot version of that helps you confirm the fix. Thanks again!

PR #251, opened 2023-12-14
Update gradle enterprise setup

ribafish:
## What's changed?

- Added Revved up by Develocity badge
- Bumped Gradle Enterprise and Common Custom User Data plugins to latest versions
- Updated remote build cache config to only push when authenticated

## What's your motivation?
- Bumped Gradle Enterprise and Common Custom User Data plugins to latest versions -> improved scans and support for gradle enterprise (server) 2023.4
- Updated remote build cache config to only push when authenticated -> if an unauthenticated push is attempted remote build cache will get turned off for the rest of the build. This can happen (does) on forked PR builds which run in an unsecure environment. 

In general, we strongly suggest to apply these fixes/updates to the rest of your projects, especially if you have a nice rewrite rule to do it for you 😄


timtebeek (approved):
Thanks a lot for helping out!

timtebeek:
I think we should be able to roll out this change through the platform using this recipe
```
type: specs.openrewrite.org/v1beta/recipe
name: org.openrewrite.UpgradeDevelocityGradlePlugins
displayName: Upgrade Develocity Gradle plugins
description: Upgrade the Develocity Gradle plugins.
recipeList:
  - org.openrewrite.gradle.plugins.UpgradePluginVersion:
      pluginIdPattern: com.gradle.enterprise
      newVersion: "3.16"
  - org.openrewrite.gradle.plugins.UpgradePluginVersion:
      pluginIdPattern: com.gradle.common-custom-user-data-gradle-plugin
      newVersion: 1.12.1
```
As per https://gradle.com/blog/develocity-global-implementation-best-practice-using-openrewrite-to-enable-develocity-project-integration-and-rollout-at-scale/ :)

PR #254, opened 2024-01-03
Add rewrite-java-21 to resolved dependencies

timtebeek:
Noticed while exploring another issue related to a project without repositories on the main project, which lead to
> org.gradle.internal.resolve.ModuleVersionNotFoundException: Cannot resolve external dependency org.openrewrite:rewrite-core:8.11.5 because no repositories are defined

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
@@ -66,6 +66,7 @@ public class ResolveRewriteDependenciesTask extends DefaultTask {
                     deps.create("org.openrewrite:rewrite-json:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-kotlin:" + extension.getRewriteKotlinVersion()),
                     deps.create("org.openrewrite:rewrite-java:" + rewriteVersion),
+                    deps.create("org.openrewrite:rewrite-java-21:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-java-17:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-java-11:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-java-8:" + rewriteVersion),


shanman190 (approved):

plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
```diff
@@ -66,6 +66,7 @@ public Set<File> getResolvedDependencies() {
                     deps.create("org.openrewrite:rewrite-json:" + rewriteVersion),
                     deps.create("org.openrewrite:rewrite-kotlin:" + extension.getRewriteKotlinVersion()),
                     deps.create("org.openrewrite:rewrite-java:" + rewriteVersion),
+                    deps.create("org.openrewrite:rewrite-java-21:" + rewriteVersion),
```
timtebeek, line 69: The first rewrite version to have rewrite-java-21 is 8.8.0.
Older rewrite versions will now fail to resolve dependencies.
Figured that's ok, but though to point it out as it failed a test.

shanman190: I agree that's probably fine. If we ever run into a case where it becomes more difficult, we can always wrap a condition around the version so as to only enable it when it's new enough.

timtebeek: We could add a conditional by looking at the rewriteVersion, but [8.8.0 was released in October](https://github.com/openrewrite/rewrite/releases/tag/v8.8.0) and there's likely other incompatibilities between older rewrite versions and the latest gradle plugin version by now anyway, so I figure there's less of a need to handle this differently.

shanman190: Yep, I wasn't meaning to add it now because like you said, it's been out for a while now and there are most certainly going to be other quirks. But more just figured that I'd mention it for the hypothetical of us being on top of the Java 25 LTS more quickly.

timtebeek: Agreed; your first comment only showed after I had replied with my second comment; so it wasn't as much a reply as a parallel similar thought. Thanks for the help and quick review!

Issue #252, opened 2023-12-29
Gradle parser does not parse subproject resources


travisboettcher:
When using the `org.openrewrite.properties.AddProperty` recipe, I noticed that the file I was trying to target (`hibernate.properties`) was not being updated. In trying to get it to work, I've seen the recipe work on other `properties` files, so I don't believe it's the recipe.

I'm wondering it the culprit is [here](https://github.com/openrewrite/rewrite-gradle-plugin/blob/ce73e203a0be839191635063b4f4e15fba69fc34/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java#L811C90-L811C90) when it is parsing the resources of a project:
```groovy
for (File resourcesDir : sourceSet.getResources().getSourceDirectories()) {
    if (resourcesDir.exists()) {
        OmniParser omniParser = omniParser(alreadyParsed);
        List<Path> accepted = omniParser.acceptedPaths(resourcesDir.toPath());
        sourceSetSourceFiles = Stream.concat(
                sourceSetSourceFiles,
                omniParser.parse(accepted, resourcesDir.toPath(), new InMemoryExecutionContext())
        );
        sourceSetSize += accepted.size();
    }
}
```

In particular, passing a new `InMemoryExecutionContext` into the omniparser rather than the existing `ctx`.

## What version of OpenRewrite are you using?

I am using the OpenRewrite Gradle plugin v6.6.2 (I also tried 6.6.3 just in case)

## How are you running OpenRewrite?

I am using the Gradle plugin and it is a multi-module project.

## What is the smallest, simplest way to reproduce the problem?

Reproduced here: [OpenRewrite-bug.zip](https://github.com/openrewrite/rewrite-gradle-plugin/files/13795580/OpenRewrite-bug.zip)

## What did you expect to see?

The file `example/src/main/resources/hibernate.properties` should be updated to include a new property, `hibernate.type.preferred_uuid_jdbc_type=CHAR`

## What did you see instead?

The file is not updated, although if I move it to the top-level directory, it does get updated. If I remove the precondition, the `gradle/gradle-wrapper.properties` file gets updated - so I don't believe it's the recipe.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Possibly!

PR #257, opened 2024-01-07
Fix resources exclusion in a multi-module project with git repo

radoslaw-panuszewski:
## What's changed?
Fixed the bug with resources in subproject being excluded from processing.

## What's your motivation?
I have a multi-module project like this:
```
.
├── src
│   └── main
│       └── resources
│           └── application-root.yml
└── subproject
    └── src
        └── main
            └── resources
                └── application-subproject.yml  
```
And I have a custom recipe to modify some YAML. While the `application-root.yml` is correctly modified, the `application-subproject.yml` is skipped. It is reproduced in the `resources in subproject committed to git are correctly processed` test case.

There is already a test case `rewriteRun applies recipe to a multi-project build` which aims to verify this situation, but it doesn't check the case when Git repo is available. There is a different code path in `OmniParser` for such case and it's not reached.

I believe that I've found a bug in the code of `DefaultProjectParser`. When the `OmniParser` is created for the root project, it is configured to exclude contents of its subprojects. But when it's created for the subproject, it's still configured to exclude the root project's subprojects, so also the subproject we want to parse.

I've added the `Project` parameter to the `omniParser()`. Please verify that the updated method usages are correct.

## Additional changes
There was a typo in `dependencyRepositoriesDeclaredInSettings` test case which caused the test to fail. I fixed it: `nevVersion` -> `newVersion`

### Checklist

### DIFF
--- a/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
+++ b/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
@@ -26,6 +26,8 @@ import org.junit.jupiter.api.condition.DisabledIf
 import org.junit.jupiter.api.condition.DisabledOnOs
 import org.junit.jupiter.api.condition.OS
 import org.junit.jupiter.api.io.TempDir
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 import org.openrewrite.Issue
 import java.io.File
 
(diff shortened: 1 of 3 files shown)


bgalek (approved):

timtebeek:
@travisboettcher this PR takes a slightly different approach to what you suggested in https://github.com/openrewrite/rewrite-gradle-plugin/issues/252 ; Any thoughts to share here?

plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
```diff
@@ -573,7 +635,7 @@ class RewriteRunTest : RewritePluginTest {
                 }
                 
                 dependencies {
-                    implementation("com.fasterxml.jackson.core:jackson-core:2.16.1")
+                    implementation("com.fasterxml.jackson.core:jackson-core:2.16.0")
```
timtebeek, line 638: Why this change?
```suggestion
                    implementation("com.fasterxml.jackson.core:jackson-core:2.16.1")
```

radoslaw-panuszewski: @timtebeek I believe the story is like this:
* the original test was written when the latest version of `jackson-core` was `2.16.0`. Back then it was passing only accidentally, because there was a typo in YAML recipe config (`nevVersion` instead of `newVersion`) and `UpgradeDependencyVersion` uses default of `latest.release` when the `newVersion` is not provided
* newer version of `jackson-core` was released (`2.16.1`) which caused the test to fail
* I noticed the test failing locally and I fixed the typo `nevVersion` -> `newVersion`
* on master it was "fixed" by changing `2.16.0` to `2.16.1` in the assertion
* I rebased on master and changed it back to `2.16.0` because it's the correct value expected by this test 😉

timtebeek: Thanks for the explanation! Makes sense now indeed, and thanks for the fix

timtebeek (approved):
Thanks a lot for the detailed look & fix!

PR #259, opened 2024-01-08
Resolve rewrite dependencies from `-Pmoderne.gradle.classpath` if provided

timtebeek:
## What's changed?
If the gradle project has no repositories defined, then resolve released Rewrite dependencies from Maven Central, and snapshots from Sonatype OSS>

## What's your motivation?
Up to now we had been failing with
> ModuleVersionNotFoundException: Cannot resolve external dependency org.openrewrite:rewrite-core:8.11.5 because no repositories are defined.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/ResolveRewriteDependenciesTask.java
@@ -60,13 +60,8 @@ public class ResolveRewriteDependenciesTask extends DefaultTask {
             if (project.getRepositories().isEmpty()) {
                 project.getRepositories().mavenCentral();
                 if (rewriteVersion.endsWith("-SNAPSHOT")) {
-                    project.getRepositories().maven(mavenArtifactRepository -> {
-                        mavenArtifactRepository.setUrl("https://oss.sonatype.org/content/repositories/snapshots/");
-                        mavenArtifactRepository.mavenContent(mavenContent -> {
-                            mavenContent.includeGroup("org.openrewrite");
-                            mavenContent.includeGroup("org.openrewrite.gradle.tooling");
-                        });
-                    });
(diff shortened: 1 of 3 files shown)


timtebeek:
Adding note here as I'm stepping away for dinner

> This might work. Though it may be better done as an afterEvaluate in the plugin class itself as I'm not sure this will be allowed in all versions (ie. Modifying the repositories during the execution phase). The other facet is that configuration cache disallows access to the project within tasks, so this would result in drifting further away from being able to support that feature (this isn't to say that there isn't a lot of existing work necessary to get to a point where it can be supported in the first place).

timtebeek:
Peter shared some insights that we're likely applying the plugin to unsuitable root projects, whereas we might instead need to apply to projects that use the Java or Kotlin plugin, as those are more likely to have repositories configured. Did I summarize that correctly @pstreef ?

pstreef:
The problem is that gradle projects are not guaranteed to have repositories defined (there is no default and not every project is actually a project with code in it).
We assume that the rootProject does have the right configuration for us to add dependencies to. Which is not always the case. And even if it does it might not have the repositories configured.

We can reproduce this issue on moderne-cli using:

```
export MODERNE_GRADLE_PLUGIN=io.moderne:modgradle:2.0.5; ./gradlew moderneLst --init-script=./mod/src/main/resources/init-mod.gradle -Pmoderne.lstOutputDirectory=/tmp/lstout
```

and we can make this work (at least further than this issue) by adding the convention plugin to the root project:
```
plugins{
    ...
    id("moderne-cli.java-conventions")
}

sambsnyd (approved):

timtebeek:
We've opted for an alternative that does not require further changes in the plugin beyond https://github.com/openrewrite/rewrite-gradle-plugin/compare/55a2219d47b68918a2e1ee1e4edc104982e86264...878e4a6431015e3a135ee5d1102d798f9d1cf41d

PR #270, opened 2024-01-24
Include compile classpath in dependencies

knutwannheden:
This commit reverts 8868dcff30f63511725961220e977d3d985d134c again. The compile classpath entries can contain important dependencies like Lombok, which typically don't exist on the runtime classpath.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -686,11 +686,13 @@ public class DefaultProjectParser implements GradleProjectParser {
                         .filter(it -> it.toString().endsWith(".java") && !alreadyParsed.contains(it))
                         .collect(toList());
 
-                // The compilation classpath doesn't include the transitive dependencies
-                // So we use the runtime classpath to get complete type information
+                // The compilation classpath is required for things like Lombok which has
+                // annotations with source retention only.
                 List<Path> dependencyPathsNonFinal;
                 try {
-                    dependencyPathsNonFinal = sourceSet.getRuntimeClasspath().getFiles().stream()
+                    dependencyPathsNonFinal = Stream.concat(
+                                    sourceSet.getRuntimeClasspath().getFiles().stream(),
+                                    sourceSet.getCompileClasspath().getFiles().stream())
                             .map(File::toPath)
                             .map(Path::toAbsolutePath)
                             .map(Path::normalize)
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -686,8 +686,9 @@ public class DefaultProjectParser implements GradleProjectParser {
                         .filter(it -> it.toString().endsWith(".java") && !alreadyParsed.contains(it))
                         .collect(toList());
 
-                // The compilation classpath is required for things like Lombok which has
-                // annotations with source retention only.
+                // The compilation classpath doesn't include the transitive dependencies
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -686,11 +686,13 @@ public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Exe
                         .filter(it -> it.toString().endsWith(".java") && !alreadyParsed.contains(it))
                         .collect(toList());
 
-                // The compilation classpath doesn't include the transitive dependencies
-                // So we use the runtime classpath to get complete type information
+                // The compilation classpath is required for things like Lombok which has
+                // annotations with source retention only.
```
sambsnyd, line 690:
```suggestion
                // The compilation classpath doesn't include the transitive dependencies
                // The runtime classpath doesn't include compile only dependencies, e.g.: lombok, servlet-api
                // So we use both together to get comprehensive type information
```

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -687,10 +687,13 @@ public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Exe
...
                 // The compilation classpath doesn't include the transitive dependencies
-                // So we use the runtime classpath to get complete type information
+                // The runtime classpath doesn't include compile only dependencies, e.g.: lombok, servlet-api
+                // So we use both together to get comprehensive type information
                 List<Path> dependencyPathsNonFinal;
                 try {
-                    dependencyPathsNonFinal = sourceSet.getRuntimeClasspath().getFiles().stream()
+                    dependencyPathsNonFinal = Stream.concat(
+                                    sourceSet.getRuntimeClasspath().getFiles().stream(),
+                                    sourceSet.getCompileClasspath().getFiles().stream())
```
shanman190, line 696: I do wonder if having the runtime classpath is accurate for the parsers. For the `compileJava` task as an example, it only receives the `compileClasspath` configuration.

knutwannheden: What you are saying makes sense to me. I found this when I was trying to track down why we no longer had any type attribution for Lombok annotations.

As a first fix it seems less drastic to revert to using compile + runtime classpath, but as you say, the compiler should probably only need the compile classpath, so I think we should definitely look into this as a follow-up.

shanman190: That sounds fine to me then. :)

shanman190 (approved):

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
knutwannheden: Please also note that the runtime classpath includes entries for `build/classes/java/main`, `build/classes/kotlin/main`, etc. This is potentially important for cross-language type attribution. Something to be investigated.

shanman190: That should be handled on the `compileClasspath` as well for the plugins that support cross source set configurations or where users have configured it already. Otherwise, the types wouldn't be able to be resolved by that particular language compiler (javac, groovyc, kotlinc, etc). Most of them don't allow full cross compilation except for the groovyc compiler which is the only one that will compile both .java and .groovy files directly.

knutwannheden: Interesting. Then this approach with only the compiler classpath sounds very compelling.

PR #273, opened 2024-01-28
Add test for issue 261

koppor:
This adds two small test for issue openrewrite/rewrite#4054.

This is just an outline.

It resembles the MWE at https://github.com/koppor/mwe-openrewrite-rewrite-gradle-plugin-issue-261/.

**However,** here, all `.java` types can be determined. (Independent of a rewrite.yml is used or not)

The test fail, because type information in `build.gradle` is missing.

This PR is only made to enable discussions of debuggin the issue.

### DIFF
--- a/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteDryRunTest.kt
+++ b/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteDryRunTest.kt
@@ -126,7 +126,7 @@ class RewriteDryRunTest : RewritePluginTest {
     }
 
     @Test
-    fun `All types can be determined with rewrite dot yaml`(
+    fun `All types can be determined with module_info_java`(
             @TempDir projectDir: File
     ) {
         gradleProject(projectDir) {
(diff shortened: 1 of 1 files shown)


koppor:
MWE (still) only with Java 11, because I wanted to keep things going (and not working on backwards compatbility)

When `module-info.java` is present, types cannot be determined. Shown by the test case.

Example file:

```java
            module demo {
                requires org.slf4j;
            }
```

koppor:
I checked the required JDK (https://github.com/openrewrite/rewrite-gradle-plugin/blob/0029c7afd15730c1679e2006784feec118c7e373/plugin/build.gradle.kts#L72). Since the plugin is build with Java 17, the check is IMHO finished. I put this PR as "ready". -- OK, to solve the issue with `module-info.java` is still open (and much more work), this is far beyond my skills.

koppor:
Current state: Issue at JavaParser. See https://github.com/openrewrite/rewrite/issues/4054#issuecomment-1966527187.

timtebeek:
Let's continue to track proper support in
- https://github.com/openrewrite/rewrite/issues/4054

PR #286, opened 2024-03-06
Fix for Java source file encoding different from UTF-8

### LINKED ISSUE #285: rewriteRun does not respect the source file charset (file.encoding / compileOptions.getEncoding())

I am using

- OpenRewrite v8.17.1
- Gradle plugin v6.8.4
- rewrite-java v8.17.1

I'm using the Gradle Plugin.

Use a project with source files that are `ISO-8859-1` encoded and contains some Latin-1 characters:
```java
/**
 * Üäé
 */
class A {
}
```

The file stays intact with `ISO-8859-1` encoding.

## What did you see instead?

```java
/**
 * ���
 */
class A {
}
```

## What is the full stack trace of any errors you encountered?

There's no error

I can offer a draft MR with a quick-fix.

renegrob:
Fixes #285 

## What's changed?

rewriteRun now respects the file encoding of the Gradle Java Plugin

### Checklist

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -30,6 +30,7 @@ import org.gradle.api.logging.Logging;
 import org.gradle.api.plugins.GroovyPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.compile.CompileOptions;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.internal.service.ServiceRegistry;
 import org.gradle.invocation.DefaultGradle;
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks a lot for the detailed fix @renegrob ! Review took a little longer with my holiday, but this looks good to me. I take it you already tried it out on projects locally?

knutwannheden:
We might have to do some digging or consult with @shanman190 but I think there is an `encoding` option on the compile task that we should check. The question is just qhen this was added. https://docs.gradle.org/current/dsl/org.gradle.api.tasks.compile.CompileOptions.html#org.gradle.api.tasks.compile.CompileOptions:encoding

shanman190:
@knutwannheden, encoding on CompileOptions has been available since at least Gradle 1.0, I'm not seeing it available in previous versions, but there was a lot of change early on. So for our use case here, it's safe to use at least from the Java compile tasks.

https://docs.gradle.org/1.0/javadoc/org/gradle/api/tasks/compile/CompileOptions.html

shanman190:
@knutwannheden, just to note that @renegrob is already pulling the encoding from the options in this PR as well in case you missed it.

renegrob:
@timtebeek Yes, I'm already using a patched version internally and it is working fine.

timtebeek:
Sounds like we're good to go then; thanks a lot @renegrob for the implementations, and @shanman190 & @knutwannheden for additional review.

PR #291, opened 2024-03-14
[WIP] Read android sources

wbonnefond:
This is a very rough WIP of attempting to read sources from Android modules.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -115,7 +115,7 @@ dependencies {
     "rewriteDependencies"("com.fasterxml.jackson.module:jackson-module-kotlin:latest.release")
 
     implementation(platform("org.openrewrite:rewrite-bom:$latest"))
-    implementation("com.android.tools.build:gradle:8.1.4")
+    compileOnly("com.android.tools.build:gradle:8.1.4")
     compileOnly("org.openrewrite:rewrite-core")
     compileOnly("org.openrewrite:rewrite-gradle")
     compileOnly("org.openrewrite.gradle.tooling:model:$latest")
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -28,6 +28,8 @@ gradlePlugin {
 }
 
 repositories {
+    gradlePluginPortal()
+    google()
     if (!project.hasProperty("releasing")) {
         mavenLocal {
             mavenContent {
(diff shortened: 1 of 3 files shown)


plugin/build.gradle.kts
```diff
@@ -74,15 +76,15 @@ java {
 }
 
 tasks.withType<JavaCompile>().configureEach {
-    options.release.set(8)
+    options.release.set(11)
```
wbonnefond, line 79: So we can add AGP and get sourcesets defined by the plugin. AGP can't compile against java 8. Would need to rework this to probably use reflection for android projects if we don't want to bump version/include AGP.

plugin/src/main/java/org/openrewrite/gradle/isolated/AndroidParser.kt
```diff
@@ -0,0 +1,178 @@
...
+    fun parseAndroidSources(
+        subproject: Project,
+        exclusions: Collection<PathMatcher>,
+        alreadyParsed: MutableSet<Path>,
+        ctx: ExecutionContext
+    ): Pair<Stream<SourceFile>, Int> {
+
+        val variants = subproject.extensions.findByType(BaseExtension::class.java)?.variants ?: return Stream.empty<SourceFile>() to 0
+
+        val sources = variants.map { variant ->
```
wbonnefond, line 71: Each gradle module can have multiple variants, and each variant can have multiple sourcesets. `variant.sourceSets` is not a list of `org.gradle.api.tasks.SourceSet`'s, so the existing parsing code can't be shared. It is a list of [SourceProvider's](https://android.googlesource.com/platform/tools/build/+/ef84df96b7b1a26a96a37506d95e3d427ee830ed/builder-model/src/main/java/com/android/builder/model/SourceProvider.java)

plugin/build.gradle.kts
timtebeek: Indeed let's not bump the required version, as that affects all our users, and there's a lot of code using Java 8 that still needs to be converted. For that same reason we still support Gradle 4.10, as much as we're aware of benefits to future versions.

timtebeek:
Thanks for the start & nudges here @wbonnefond ! Bryce has taken on this effort and will continue to work in
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/327

timtebeek:
Hi @wbonnefond ; You'll likely have noticed https://github.com/openrewrite/rewrite-gradle-plugin/pull/327 was merged. 🎉 
We're thinking of starting a `rewrite-android` recipe module to facilitate collaboration on recipes for Android. Did you already have recipes in mind & want to get involved?

PR #294, opened 2024-03-22
Sum estimate time saved and log it

nicolasb29:
[#296](https://github.com/openrewrite/rewrite-gradle-plugin/issues/296)

## What's changed?
After log changes for each result, sum estimate time saved and log it.

## What's your motivation?
OpenRewrite is working fine and it is very helpful. To show the importance and the time saved, I think it's interesting to show the estimate time saved.

### Checklist
- [X ] I've added unit tests to cover both positive and negative cases
- [X ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X ] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -30,6 +30,7 @@ import org.gradle.api.logging.Logging;
 import org.gradle.api.plugins.GroovyPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.compile.CompileOptions;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.internal.service.ServiceRegistry;
 import org.gradle.invocation.DefaultGradle;
@@ -98,6 +99,7 @@ import static java.util.Collections.*;
 import static java.util.stream.Collectors.*;
 import static org.openrewrite.PathUtils.separatorsToUnix;
 import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.tree.ParsingExecutionContextView.view;
 
 public class DefaultProjectParser implements GradleProjectParser {
     private static final String LOG_INDENT_INCREMENT = "    ";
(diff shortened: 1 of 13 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -439,17 +449,21 @@ public void run(ResultsContainer results, ExecutionContext ctx) {
...
                     logger.lifecycle("Changes have been made to " +
                                      result.getBefore().getSourcePath() +
                                      " by:");
                     logRecipesThatMadeChanges(result);
+                    estimateTimeSaved = estimateTimeSavedSum(result, estimateTimeSaved);
                 }
 
                 logger.lifecycle("Please review and commit the results.");
 
+                logger.lifecycle("Estimate time saved : {}", estimateTimeSaved);
```
timtebeek, line 465: By default this will print something like `PT8H6M12.345S` which is perhaps not very user friendly. Would we want to make this more human readable?

timtebeek (commented):
I like the suggestion! I've proposed it internally to see if we want this on both of the OSS plugins, before I go and merge it here.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
timtebeek: I think I like this quick no-library solution to a more readable form best: https://stackoverflow.com/a/40487511

nicolasb29: I made a function which is using string builder but I can use the way in your link.

nicolasb29: There is common library between Gradle and Maven plugins ?

timtebeek: Let's opt for that four liner in the link above; that way it's small enough to not have to share that between the two plugins.

nicolasb29: It is on the pull request.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -440,17 +458,21 @@ public void run(ResultsContainer results, ExecutionContext ctx) {
...
                     logger.lifecycle("Changes have been made to " +
                                      result.getBefore().getSourcePath() +
                                      " by:");
                     logRecipesThatMadeChanges(result);
+                    estimateTimeSaved = estimateTimeSavedSum(result, estimateTimeSaved);
                 }
 
                 logger.lifecycle("Please review and commit the results.");
 
+                logger.lifecycle("Estimate time saved : {}", formatDuration(estimateTimeSaved));
```
timtebeek, line 474:
```suggestion
                logger.lifecycle("Estimate time saved: {}", formatDuration(estimateTimeSaved));
```

timtebeek (approved):
Thanks a lot! Really good to show more folks how much time is saved. :)

timtebeek:
Awesome work; I've logged an issue to remind myself we'll want the same for Maven
- https://github.com/openrewrite/rewrite-maven-plugin/issues/781

PR #295, opened 2024-03-22
refactor: Update Gradle wrapper

shanman190:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.gradle.UpdateGradleWrapper?organizationId=T3BlblJld3JpdGU%3D


knutwannheden (approved):

sambsnyd (approved):

shanman190:
The `UserInputHandler` class has had some modifications for Gradle 8.7 as it appears the Gradle team is beginning to clean up the API a bit in an effort to make a public API version of the feature available. This will presently require a version check to make the OpenRewrite plugin compatible with the new version.

timtebeek:
Adding the details here to save others a click:
```
Compiling with JDK Java compiler API.
/home/runner/work/rewrite-gradle-plugin/rewrite-gradle-plugin/plugin/src/main/java/org/openrewrite/gradle/isolated/ui/RecipeDescriptorTreePrompter.java:62: error: cannot find symbol
                prompted = prompter.askQuestion(query.toString(), "");
                                   ^
  symbol:   method askQuestion(String,String)
  location: variable prompter of type UserInputHandler
```
https://github.com/openrewrite/rewrite-gradle-plugin/blob/6121f79f122f69f09457734ee9fd28a5888493db/plugin/src/main/java/org/openrewrite/gradle/isolated/ui/RecipeDescriptorTreePrompter.java#L59-L62

shanman190:
I'm honestly kinda surprised that more folks haven't reported this issue. It kinda makes me think that the `rewriteDiscover` task tends to not be utilized at all.

timtebeek:
Yes I'm thinking the same: that task and the equivalent in the Maven plugin are likely not used much, if at all. They can be useful when troubleshooting I suppose, but there's less of a need for that now that we suggest the closest match too
- https://github.com/openrewrite/rewrite/pull/4169

We could remove these if we don't feel like maintaining these goals going forward. I doubt a knowledgeable user would need them, nor if a novice use would find them.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
-distributionSha256Sum=9631d53cf3e74bfa726893aee1f8994fee4e060c401335946dba2156f440f24c
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+distributionSha256Sum=544c35d6bd849ae8a5ed0bcea39ba677dc40f49df7d1835561582da2009b961d
```
timtebeek, line 4:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
distributionSha256Sum=d725d707bfabd4dfdc958c624003b3c80accc03f7037b5122c4b1d0ef15cecab
```

timtebeek (approved):

PR #281, opened 2024-02-29
Switch to JavaPluginExtension

timtebeek:
As discussed replicating the same change here ; but I'm getting failures on Gradle 4.10.
```
java.lang.NoSuchMethodError: org.gradle.api.plugins.JavaPluginExtension.getSourceSets()Lorg/gradle/api/tasks/SourceSetContainer;
	at org.openrewrite.gradle.RewritePlugin.lambda$configureProject$5(RewritePlugin.java:101)
	at org.gradle.api.internal.DefaultDomainObjectCollection.all(DefaultDomainObjectCollection.java:158)
	at org.openrewrite.gradle.RewritePlugin.configureProject(RewritePlugin.java:87)
	at org.openrewrite.gradle.RewritePlugin.lambda$apply$0(RewritePlugin.java:78)
	at org.gradle.api.internal.project.BuildOperationCrossProjectConfigurator$4.execute(BuildOperationCrossProjectConfigurator.java:108)
	at org.gradle.internal.Actions.with(Actions.java:245)
```

PR #297, opened 2024-03-26
Stop using deprecated Project.getBuildDir()

quaff:
use `Project.getLayout().getBuildDirectory()` instead

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewriteDryRunTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewriteDryRunTask.java
@@ -32,7 +32,7 @@ public class RewriteDryRunTask extends AbstractRewriteTask {
     // On Gradle 4.0 annotating something returning a Path with @OutputFile triggers a bug that deadlocks Gradle
     @OutputFile
     public File getReportPath() {
-        return getProject().getBuildDir().toPath().resolve("reports").resolve("rewrite").resolve("rewrite.patch").toFile();
+        return getProject().getLayout().getBuildDirectory().get().getAsFile().toPath().resolve("reports").resolve("rewrite").resolve("rewrite.patch").toFile();
     }
 
     @Inject
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -302,8 +302,7 @@ public class DefaultProjectParser implements GradleProjectParser {
                 heapMetrics.bindTo(meterRegistry);
                 new JvmMemoryMetrics().bindTo(meterRegistry);
 
-                //noinspection deprecation
-                File rewriteBuildDir = new File(project.getBuildDir(), "/rewrite");
+                File rewriteBuildDir = project.getLayout().getBuildDirectory().dir("rewrite").get().getAsFile();
                 if (rewriteBuildDir.exists() || rewriteBuildDir.mkdirs()) {
                     File rewriteGcLog = new File(rewriteBuildDir, "rewrite-gc.csv");
                     try (FileOutputStream fos = new FileOutputStream(rewriteGcLog, false);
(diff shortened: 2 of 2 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -99,9 +99,8 @@ private static void configureProject(Project project, RewriteExtension extension
 
             //Collect Java metadata for each project (used for Java Provenance)
             //Using the older javaConvention because we need to support older versions of gradle.
-            @SuppressWarnings("deprecation")
-            JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention.class);
-            javaConvention.getSourceSets().all(sourceSet -> {
+            JavaPluginExtension javaPluginExtension = project.getExtensions().getByType(JavaPluginExtension.class);
+            javaPluginExtension.getSourceSets().all(sourceSet -> {
```
knutwannheden, line 103: We tried this in the past, but it looks like `getSourceSets()` was not added until 7.1 so we reverted the change again: https://github.com/openrewrite/rewrite-gradle-plugin/pull/281#issuecomment-1970694368

However, we will also want to be Gradle 9 compatible, so we might want to use both APIs here, depending on the Gradle version.

What do you think @shanman190 ?

shanman190: Yeah, that'll have to be what we do. What I would probably do is hide the complexity in a `SourceSetContainer getSourceSets()` method. For pre-7.1, we'd use reflection to load the convention class, get the instance, then get the `SourceSetContainer` off of it. Then for 7.1 and later, we'd just code directly to the `JavaPluginExtension` directly. This would allow us to remain backwards compatible as well as being forward compatible to whenever Gradle finally decides to remove the `JavaPluginConvention` class (it's been deprecated since the 7.x line where it was originally scheduled to be removed at 8.0, but was delayed).

quaff: We can mitigate it by checking which version is running to decide execution of code path, but there is another problem, it breaks test `rewriteRun applies recipe to a multi-project build(File)`:
```
org.opentest4j.AssertionFailedError: 
expected: 
  "package com.foo;
  
  import org.junit.Test;
  
  public class ATestClass {
  
      @Test
      public void passes() {
      }
  }"
 but was: 
  "package com.foo;
  
  import org.junit.Test;
  
  public class ATestClass {
  
      @Test
      public void passes() { }
  }"
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at org.openrewrite.gradle.RewriteRunTest.rewriteRun applies recipe to a multi-project build(RewriteRunTest.kt:255)
```

quaff:
I reverted changes about `JavaPluginConvention`.

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
knutwannheden: Not entirely sure what the issue is here, but I am sure that could be fixed.

knutwannheden: @quaff If you would like to contribute a PR which allows the build plugin to use `JavaPluginExtension` rather than `JavaPluginConvention` in Gradle 7+ environments, that would definitely be most welcome ❤️

quaff: I will give a shot if I'm spare, but I think this PR should focus on `Project.getBuildDir()`.

knutwannheden:
Thanks 🙏

PR #299, opened 2024-03-27
Stop using JavaPluginConvention for Gradle 7.1+

### LINKED ISSUE #307: Gradle 8.x Deprecates PluginConvention and will be Removed in Gradle 9.0

This is used in https://github.com/openrewrite/rewrite-gradle-plugin/blob/main/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java#L103
See 
https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecated_access_to_conventions

quaff:
The org.gradle.api.plugins.JavaPluginConvention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.7/userguide/upgrading_version_8.html#java_convention_deprecation

Fixes #307

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -21,8 +21,10 @@ import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.plugins.JavaBasePlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.plugins.JavaPluginExtension;
 import org.gradle.api.plugins.quality.CheckstyleExtension;
 import org.gradle.api.plugins.quality.CheckstylePlugin;
+import org.gradle.api.tasks.SourceSetContainer;
 
 import java.io.File;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -98,10 +100,16 @@ private static void configureProject(Project project, RewriteExtension extension
             }
 
             //Collect Java metadata for each project (used for Java Provenance)
-            //Using the older javaConvention because we need to support older versions of gradle.
-            @SuppressWarnings("deprecation")
-            JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention.class);
-            javaConvention.getSourceSets().all(sourceSet -> {
+            SourceSetContainer sourceSets;
+            if (project.getGradle().getGradleVersion().compareTo("7.0") > 0) {
```
shanman190, line 104:
```suggestion
            if (GradleVersion.current().compareTo(GradleVersion.version("7.1")) >= 0) {
```

Specifically, the `GradleVersion` has better semver semantics and there would have been bugs introduced with the previous implementation for 7.0.1 and 7.0.2 as the `JavaPluginExtension#getSourceSets()` method is first introduced in 7.1.

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -98,10 +100,16 @@ private static void configureProject(Project project, RewriteExtension extension
...
+                sourceSets = project.getExtensions().getByType(JavaPluginExtension.class).getSourceSets();
+            } else {
+                //Using the older javaConvention because we need to support older versions of gradle.
+                @SuppressWarnings("deprecation")
+                JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention.class);
+                sourceSets = javaConvention.getSourceSets();
```
shanman190, line 110: This will work for now, but if 9.0 actually removes the conventions or `JavaPluginConvention`, then this code will no longer compile when the Gradle wrapper is upgraded to 9.0 in this repo and will require being refactored again.

quaff: > this code will no longer compile when the Gradle wrapper is upgraded to 9.0 in this repo and will require being refactored again.

Yes, but I think released versions will still works fine with 9.0+, and no warning for 7.1+.

timtebeek:
Thanks a lot for this @quaff ! Any concerns before we merge this in @shanman190 ?

shanman190:
For now this is probably fine for now at least.

The one concern that comes to mind is on JVMs that preemptively load all classes and the `JavaPluginConvention` actually being removed. In that case, we would most definitely begin to receive reports of the plugin failing to load the old class. These JVMs seem to be fairly uncommon though.

PR #310, opened 2024-07-19
Add ability to export dataTables

ryanwalker:
## What's changed?
Added the ability to export data tables.

## What's your motivation?
I was reading about them in the docs and found that they were only implemented in maven and there was a request to add them to the gradle plugin. 

I did see that this open issue that talks about adding some configuration that would allow you more control over turning certain csv exports on or off. From a cursory look, it seems like I might have to change rewrite-core for that to work. I'm not opposed to doing that, but I thought I'd create this PR to get thoughts before moving too far forward.

I did not add any tests yet either. I'll be more than happy to update this to a prod-ready state.

I added to the `RewriteExtension`:

```
rewrite {
    exportDatatables = true
}

```

It defaults to false.

Here's a screenshot of the cvs file output:
image

### Checklist

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -89,6 +89,8 @@ import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -1255,6 +1257,14 @@ protected ResultsContainer listResults(ExecutionContext ctx) {
 
         logger.lifecycle("All sources parsed, running active recipes: {}", String.join(", ", getActiveRecipes()));
         RecipeRun recipeRun = recipe.run(new InMemoryLargeSourceSet(sourceFiles), ctx);
+
+        if (extension.isExportDatatables()) {
```
ryanwalker, line 1261: This does exactly what the maven plugin does.

timtebeek:
Thanks a lot for closing the gap here. Indeed seems like a mostly straightforward add with most functionality in core. I'll take a closer look after the weekend, but suspect that we should be able to add this pretty soon next week

PR #322, opened 2024-08-15
Do not add Kotlin and Groovy sources from `build/` to LST

timtebeek:
## What's changed?
Reuse the mechanism we added for `protos/build/generated` for `build/generated-sources`.

## What's your motivation?
Files like `build-src/build/generated-sources` are parsed into the LST, and modified by recipes. That last part is undesirable, as the diffs would then target files not part of the git repository, and fail to apply from a dryRun fix.patch file.

n | sourcePath | sourceFileType | buildTool | buildToolVersion | checksum | lineCount
-- | -- | -- | -- | -- | -- | --
1 | build-src/build/generated-sources/kotlin-dsl-external-plugin-spec-builders/kotlin/gradle/kotlin/dsl/plugins/_12895b5fa930feaf40eec89923f04a18/PluginSpecBuilders.kt | K.CompilationUnit | modgradle | 3.17.0 | 55d3a6f819a2d911e8dc50fcdba328c6 | 1186
2 | build-src/build/generated-sources/kotlin-dsl-accessors/kotlin/gradle/kotlin/dsl/accessors/_f68e08f112e4a379392eaca54506dc98/Accessorsct2ylp5ktt4rxe14s34wgi9de.kt | K.CompilationUnit | modgradle | 3.17.0 | f4d3580d463adfe088c5cc21c590086f | 76
3 | build-src/build/generated-sources/kotlin-dsl-accessors/kotlin/gradle/kotlin/dsl/accessors/_f68e08f112e4a379392eaca54506dc98/Accessorsd6fwhfk2ip9kv56d7w2vygrh6.kt | K.CompilationUnit | modgradle | 3.17.0 | 1a203f1b9aaeb9223e2565c7cd25fa37 | 76

There's likely a better way, but I'm having a bit of trouble working that out. Figured open this PR to start the discussion.

What I'd done previously in an attempt to exclude these files is add them to Git ignore:
https://github.com/openrewrite/rewrite-recipe-bom/commit/9f5840cc1d8d8b4f4795546b95c8242622d6c116
But somehow they are still picked up; I had thought we already excluded files ignored in git.

## Any additional context
I know in the Maven plugin we explicitly exclude generated sources, but I'm not immediately seeing an equivalent here.
https://github.com/ammachado/rewrite-maven-plugin/blob/f900a9a6365a4d02c0083dd29e14927d6ffa6576/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L425-L430

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -147,6 +147,7 @@ public class DefaultProjectParser implements GradleProjectParser {
     }
 
     private static final Map<Path, GitProvenance> REPO_ROOT_TO_PROVENANCE = new HashMap<>();
+
     private @Nullable GitProvenance gitProvenance(Path baseDir, @Nullable BuildEnvironment buildEnvironment) {
         try {
             // Computing git provenance can be expensive for repositories with many commits, ensure we do it only once per build
@@ -660,13 +661,14 @@ public class DefaultProjectParser implements GradleProjectParser {
             }
 
             if (subproject.getPlugins().hasPlugin("org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension") ||
-                subproject.getExtensions().findByName("kotlin") != null && subproject.getExtensions().getByName("kotlin").getClass()
-                        .getCanonicalName().startsWith("org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension")) {
+                    subproject.getExtensions().findByName("kotlin") != null && subproject.getExtensions().getByName("kotlin").getClass()
+                            .getCanonicalName().startsWith("org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension")) {
                 sourceFileStream = sourceFileStream.concat(parseMultiplatformKotlinProject(subproject, exclusions, alreadyParsed, ctx));
             }
 
             Charset sourceCharset = Charset.forName(System.getProperty("file.encoding", "UTF-8"));
 
+            Path buildDirPath = baseDir.relativize(subproject.getLayout().getBuildDirectory().get().getAsFile().toPath());
             for (SourceSet sourceSet : sourceSets) {
                 Stream<SourceFile> sourceSetSourceFiles = Stream.of();
                 int sourceSetSize = 0;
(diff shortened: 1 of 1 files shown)


shanman190:
I suspect that we're actually including all types of generated sources that are produced by various plugins as they will appear within the source files from `sourceSet.getAllSources()`. I think the current proto implementation was a bandaid around the problem, but never actually addressed it. 

What should probably be being done is a filter operation when collecting all of the unparsed sources and excluding them when they are contained within `subproject.getLayout().getBuildDirectory()`. This would be directly akin to what was done within the Maven plugin.

timtebeek:
Thanks for the hint! Noticed we only had that pattern for Java; now also for Kotlin & Groovy + Kotlin multi platform. Should help I think, although it's a bit cumbersome to test.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -737,7 +739,7 @@ public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Exe
                             .flatMap(jp -> jp.parse(javaPaths, baseDir, ctx))
                             .map(cu -> {
                                 if (isExcluded(exclusions, cu.getSourcePath()) ||
-                                    cu.getSourcePath().startsWith(baseDir.relativize(subproject.getLayout().getBuildDirectory().get().getAsFile().toPath()))) {
+                                        cu.getSourcePath().startsWith(buildDirPath)) {
```
shanman190, line 742: Do we want to even be parsing these files at all with OpenRewrite?

shanman190 (approved):

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
timtebeek: I thought we did so for the type attribution; that's at least [the comment over on the Maven plugin](https://github.com/openrewrite/rewrite-maven-plugin/blob/ecf91b17f3dfe11cf35fa10e8c448651654497e5/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L387-L393). Hadn't thought to change that here just yet. Just excluding them after they are parsed to prevent them ending up in the LST themselves.

shanman190: So we have so far, yeah, but we also trigger the compile phases in Gradle as well prior to the rewrite tasks, so technically we could put the source set's own output classes onto the parser's classpath to still get the necessary type attribution and bypass parsing the generated types.

Just a thought though.

timtebeek: Could indeed be a nice improvement to stop ourselves from parsing source files whose LSTs we'll discard anyway, or perhaps even having two sources of the same type information; compiled vs source.

Issue #323, opened 2024-08-16
gradleParser causes classloading problems when combined with the plugin transformation


blipper:
## How are you running OpenRewrite?
Via custom Gradle tasks and isolated classloaders.

## What is the smallest, simplest way to reproduce the problem?
The problem is that rewrite-java/src/main/java/org/openrewrite/java/RemoveUnusedImports.java is a recipe but lives in the same JAR as 

We override the getResolvedDependencies but this needs is from another classloader so it blows up
https://github.com/openrewrite/rewrite-gradle-plugin/blob/main/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java#L150

Can we move all of these java recipes to their own JAR so they don't conflict with the OpenRewrite java engine itself?

## What did you see instead?

```
> loader constraint violation: when resolving method 'org.openrewrite.groovy.GroovyParser$Builder org.openrewrite.groovy.GroovyParser$Builder.typeCache(org.openrewrite.java.internal.JavaTypeCache)' the class loader org.openrewrite.gradle.RewriteClassLoader @5e6d03b4 of the current class, org/openrewrite/gradle/isolated/DefaultProjectParser, and the class loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @4a2a073c for the method's defining class, org/openrewrite/groovy/GroovyParser$Builder, have different Class objects for the type org/openrewrite/java/internal/JavaTypeCache used in the signature (org.openrewrite.gradle.isolated.DefaultProjectParser is in unnamed module of loader org.openrewrite.gradle.RewriteClassLoader @5e6d03b4, parent loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @4a2a073c; org.openrewrite.groovy.GroovyParser$Builder is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @4a2a073c, parent loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @78fef302)
```

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

blipper:
The problem is that my plugin is written in gradle/Groovy so the parser loaded automatically.   

[...]

PR #325, opened 2024-08-20
Don't init the gradle parser until we have confirmed the settings files is not excluded.

blipper:
## What's changed?
gradle.settings 

## What's your motivation?
Needed to work around 
- https://github.com/openrewrite/rewrite-gradle-plugin/issues/323

### Checklist


timtebeek:
Looks reasonable to me; thanks for taking this on! Any feedback from @shanman190 before we merge? Since I'd noticed the both of you discussed on the linked issue.

shanman190 (approved):
This looks like a reasonable change to me. A little surprised that's the part that was leading to the classloading issue.

PR #326, opened 2024-08-21
Update RewritePlugin to load tasks lazily

### LINKED ISSUE #288: Plugin Creates Tasks Immediately Rather Than Lazily in Configuration or Execution Phases

I am using

- OpenRewrite v2.8.0
- Maven/Gradle plugin v6.10.0

I am using the Gradle plugin, and my project is a multi module project. I am using convention plugins to distribute the plugin to all of the eligible modules in the build. I have over 500 modules (potentially) using the plugin. It is a private repository so I can't share any code.

When I add the Openrewrite plugin to the build it creates its tasks immediately rather than relying on lazy configuration. I know this via our Gradle Enterprise build scans. This significantly slows down the configuration phase of our build which has hundreds of modules and tens of thousands of tasks. I am attaching a screenshot of Gradle Enterprise that shows this

SCR-20240313-plxa

Lazy configuration has been around since Gradle 5.x and is the standard for defining tasks. I would expect tasks to be created during configuration or execution rather than creating them immediately. This is to make sure that modules that don't need those tasks don't have to wait for them to be created. 

## What did you see instead?

As the above image shows, I see the openrewrite tasks being created immediately.

I would be open to taking a stab at updating the plugin to register the tasks lazily but in order to do this support for Gradle 4.x would need to be removed. That is a decision I can't make unilaterally. Gradle 4.0 was released in June of 2017 and the last of the 4.x line was released in December of 2018. I do think it's reasonable to drop support for a version that is over five years old. This is going to be even more of a problem as Gradle evolves with version 9.x and beyond. I think it's time to bring this plugin into the modern Gradle world.
Until this plugin is more modern, I can't advise my company to use Openrewrite or Moderne. It's a nonstarter as it affects build performance too much.

mattbertolini:
## What's changed?

The RewritePlugin was changed to call the register() method instead of the create() method for the three tasks. This method takes a configuration lambda where the existing configuration can be done. Next, the tasks' usages were replaced with the equivalent TaskProvider reference. Finally, fetching the appropriate JavaCompile tasks was updated to use a lazy load method (named()), so those tasks are not eagerly created. The plugin's Gradle version was unchanged and should still support Gradle 4.10.

## What's your motivation?

- fixes #288 

The current structure of RewritePlugin eagerly loads the RewriteRun, RewriteDryRun, and RewriteDiscover tasks. This is a problem in Gradle builds with a large number of modules, as it has to create hundreds or even thousands of tasks during configuration. This will happen even if the tasks are not used in that run. This commit updates the plugin by lazily loading these tasks to avoid this issue.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -31,6 +31,7 @@ import org.gradle.api.plugins.quality.CheckstyleExtension;
 import org.gradle.api.plugins.quality.CheckstylePlugin;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.SourceSetContainer;
+import org.gradle.api.tasks.TaskProvider;
 
 import java.io.File;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


shanman190 (approved):
Thanks for the contribution, @mattbertolini!

Confirmed that all methods are available in Gradle 4.10.

timtebeek:
Thanks a lot @mattbertolini, not only for the code change, but also for the detailed background here and in #288, as well as your patience waiting for #228 to land before coming back here with the fix.

mattbertolini:
Happy to help!

PR #327, opened 2024-08-22
Add support for Android source sets

bryceatmoderne:
## What's changed?
- Updated `rewrite-gradle-plugin` to read `android` project source sets using the Android Gradle Plugin. 
- in order to avoid `ClassCastException`s when loading AGP classes using `RewriteClassLoader` the class loader of the AGP plugin instance is now passed in the `RewriteClassLoader` constructor and all `com.android...` classes are loaded using it
- all logic to read android source sets is encapsulated in `AndroidProjectParser` which is lazily allocated to ensure `rewrite-gradle-plugin` tasks do not attempt to load AGP classes for non-android projects
- some reflection was required to accommodate different types across AGP versions; for example see `AndroidProjectCompileOptions`
- these changes have been tested using AGP versions 8.6.0 to 3.3.0
- the `compileOnly` dependency of AGP v 7.0.4 was choses as it was the last version to support JDK 8 which `rewrite-gradle-plugin` targets
- note that despite a new AGP Variant API, the majority of classes and methods required to read android source sets have remained constant between versions 3.3.0 through 8.6.0 despite some being deprecated (and their documented removed from reference guides). Once google decides to remove these classes in a future AGP release, a new approach will need to be considered to support android source sets

## What's your motivation?
These changes will allow `rewrite-gradle-plugin` to run recipes on android project repositories.

I tried several approaches include adding AGP jars to `RewriteClassLoader` but could not avoid `ClassCastExceptions`s when assigned gradle decorated AGP classes to ones loaded by `RewriteClassLoader`

### Checklist


plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
```diff
@@ -0,0 +1,292 @@
...
+import java.io.File
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+
+/**
+ * Utility to help with writing android projects to disk to assist with plugin testing
+ */
+class AndroidProjectSpec(
```
bryceatmoderne, line 28: I'll probably make this a subclass of `GradleProjectSpec` and remove most of this code.

plugin/build.gradle.kts
```diff
@@ -140,6 +148,11 @@ dependencies {
     }
 }
 
+// This is necessary to add android build tools to the test runtime classpath
+tasks.withType<PluginUnderTestMetadata>().configureEach {
+    pluginClasspath.from(configurations["testDependencies"])
+}
```
shanman190, line 154: This is typically handled by having a `compileOnly` and `testRuntimeOnly` from a testing standpoint.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -636,62 +656,283 @@ public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Exe
...
+            // Some source sets get misconfigured to have the same directories as other source sets
+            // Prevent files which appear in multiple source sets from being parsed more than once
+            for (File file : sourceSet.getAllSource().getSourceDirectories().getFiles()) {
+                alreadyParsed.add(file.toPath());
+            }
+        }
+        return sourceFileStream;
+    }
+
+    private SourceFileStream parseAndroidProjectSourceSets(Project subproject,
```
shanman190, line 912: callout: Future problem will almost certainly be needing an Android specific marker to carry various information such as the min and max SDK versions, etc.

plugin/build.gradle.kts
bryceatmoderne: We use gradle's TestKit for testing plugins which doesn't use `testRuntimeOnly` when building the classpath the plugin is tested with. With the above code, you'd see something like
```
* Exception is:
java.lang.NoClassDefFoundError: com/android/build/api/dsl/CompileOptions
	at org.openrewrite.gradle.DelegatingProjectParser.<init>(DelegatingProjectParser.java:65)
	at org.openrewrite.gradle.AbstractRewriteTask.getProjectParser(AbstractRewriteTask.java:89)
```

The issue and solution is described here: https://discuss.gradle.org/t/issue-in-a-custom-plugin-with-plugins-dsl-and-android-plugin-dependency/29123

shanman190: Ahh, right, right. I've also used the method to add additional `testSourceSets` to handle that issue as well.

plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
```diff
@@ -0,0 +1,292 @@
...
+                    .groupValues[1]
+            } else {
+                ""
+            }.replace(".", "/")
+            val clazz = ".*(class|interface|enum)\\s+([a-zA-Z0-9-_]+)".toRegex(RegexOption.MULTILINE)
+                .find(groovySource)!!.groupValues[2]
+            val path = if (packageDecl.isEmpty()) {
+                "$name/groovy/$clazz.groovy"
+            } else {
+                "$name/groovy/$packageDecl/$clazz.groovy"
```
knutwannheden, line 251: You will probably want to replace the dots with slashes in the package name.

plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
```diff
@@ -0,0 +1,292 @@
...
+                    .groupValues[1]
+            } else {
+                ""
+            }.replace(".", "/")
+            val clazz = ".*(class|interface|enum)\\s+([a-zA-Z0-9-_]+)".toRegex(RegexOption.MULTILINE)
+                .find(javaSource)!!.groupValues[2]
+            val path = if (packageDecl.isEmpty()) {
+                "$name/java/$clazz.java"
+            } else {
+                "$name/java/$packageDecl/$clazz.java"
```
knutwannheden, line 211: You will probably want to replace the dots with slashes in the package name.

plugin/build.gradle.kts
bryceatmoderne: The need for this workaround was removed with the current approach of adding the plugin class loader to `RewriteClassLoader`

plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
bryceatmoderne: This class was not needed and I've deleted it.

plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
bryceatmoderne: I've deleted this class.

plugin/src/main/kotlin/org/openrewrite/gradle/AndroidProjectSpec.kt
bryceatmoderne: This class was not needed and I've deleted it.

plugin/src/main/java/org/openrewrite/gradle/isolated/AndroidProjectParser.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    SourceFileStream parseProjectSourceSets(Project project,
+                                            Path buildDir,
+                                            Charset sourceCharset,
+                                            Set<Path> alreadyParsed,
+                                            Collection<PathMatcher> exclusions,
+                                            ExecutionContext ctx,
+                                            OmniParser omniParser) {
+        // FIXME: Does this affect progress bars?
```
bryceatmoderne, line 71: Has to be resolved.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -636,238 +658,323 @@ public Stream<SourceFile> parse(Project subproject, Set<Path> alreadyParsed, Exe
...
+        }
+    }
+
+    private SourceFileStream parseGradleProjectSourceSets(Project subproject,
+                                                          Path buildDir,
+                                                          Charset sourceCharset,
+                                                          Set<Path> alreadyParsed,
+                                                          Collection<PathMatcher> exclusions,
+                                                          ExecutionContext ctx) {
+        // FIXME: Does this affect progress bars?
```
bryceatmoderne, line 739: Has to be resolved.

bryceatmoderne: Done

shanman190:
For the OSS build plugins, the plugin classloader part wouldn't be necessary. However, when testing via the tooling API or loading the plugin via the initscript, it results in a different location as to where to find plugins. To illustrate this a bit more, consider the typical classloader hierarchy below for a Gradle project:

```
JVM
  Gradle
    Initscript
      Settings
        Root Project
          Subproject A
            Subprojects A.1
              ...
            Subprojects A.2
              ...
          Subproject B
            Subproject B.1
              ...
            Subproject B.2
              ...
          ...
```

For a moment, let's just focus in on where we would find `RewriteClassLoader` and Android Gradle Plugin's `BaseAppModuleExtension` to make things simplier.

When running via the OSS build plugins applied to the root project, then both classes would appear in the "Root Project" classloader above. When they are both in "Subproject A", they would both appear within the "Subproject A" classloader. Now if the OSS build plugin is at the root, but AGP is in "Subproject A", the plugin classloader logic here is necessary because the class is in a child classloader rather than a parent classloader. When the OSS build plugin is applied via an Initscript, it's now a 4th generational parent from the AGP class.

As a result of the placement of the various jars into each of their classloaders for the necessary classpath isolation, this is what is resulting in the differing behavior and ultimately the necessity of loading the AGP classes from a particular classloader. This actually compounds with each additional plugin and is probably one case where Kotlin Multiplatform may not be working fully as an example.

bryceatmoderne:
> For the OSS build plugins, the plugin classloader part wouldn't be necessary. However, when testing via the tooling API or loading the plugin via the initscript, it results in a different location as to where to find plugins. To illustrate this a bit more, consider the typical classloader hierarchy below for a Gradle project:
> 
> ```
> JVM
>   Gradle
>     Initscript
>       Settings
>         Root Project
>           Subproject A
>             Subprojects A.1
>               ...
>             Subprojects A.2
>               ...
>           Subproject B
>             Subproject B.1
>               ...
>             Subproject B.2
>               ...
>           ...
> ```
> 
> For a moment, let's just focus in on where we would find `RewriteClassLoader` and Android Gradle Plugin's `BaseAppModuleExtension` to make things simplier.
> 
> When running via the OSS build plugins applied to the root project, then both classes would appear in the "Root Project" classloader above. When they are both in "Subproject A", they would both appear within the "Subproject A" classloader. Now if the OSS build plugin is at the root, but AGP is in "Subproject A", the plugin classloader logic here is necessary because the class is in a child classloader rather than a parent classloader. When the OSS build plugin is applied via an Initscript, it's now a 4th generational parent from the AGP class.
> 
> As a result of the placement of the various jars into each of their classloaders for the necessary classpath isolation, this is what is resulting in the differing behavior and ultimately the necessity of loading the AGP classes from a particular classloader. This actually compounds with each additional plugin and is probably one case where Kotlin Multiplatform may not be working fully as an example.

Thank you for these insights @shanman190. Much appreciated!

PR #330, opened 2024-09-12
Fail hard on failure to resolve dependencies

timtebeek:
## What's changed?
Fail hard on failure to resolve dependencies, as opposed to continuing with partial type information.

## What's your motivation?
Partial type information can lead to missing changes, and hard to trouble shoot issues. We'd rather fail fast and hard, such that the underlying issues can be addressed.

Not too familiar with Gradle plugin development; throwing a GradleException seemed the best fit to achieve something similar to [what we did in the Maven plugin](https://github.com/openrewrite/rewrite-maven-plugin/compare/f59215908381abfd4fbdea116b442f58752292a2...59404e58d219ab900ff3787497cf148f53253ac8).

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -1303,9 +1303,9 @@ public class DefaultProjectParser implements GradleProjectParser {
                 try {
                     implementationClasspath = rewriteImplementation.resolve();
                 } catch (Exception e) {
-                    logger.warn("Failed to resolve dependencies from {}:{}. Some type information may be incomplete",
-                            subproject.getPath(), implementationName);
-                    implementationClasspath = emptySet();
+                    throw new RuntimeException(String.format("Failed to resolve the dependencies from %s:%s; " +
+                                    "We can not reliably continue without this information.",
+                            subproject.getPath(), implementationName), e);
                 }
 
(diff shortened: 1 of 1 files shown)


shanman190 (approved):
LGTM!

GradleException, or any other exception, will result in the task being marked as failed.

plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -1303,9 +1305,9 @@ private SourceFileStream parseMultiplatformKotlinProject(Project subproject, Col
                 try {
                     implementationClasspath = rewriteImplementation.resolve();
                 } catch (Exception e) {
-                    logger.warn("Failed to resolve dependencies from {}:{}. Some type information may be incomplete",
-                            subproject.getPath(), implementationName);
-                    implementationClasspath = emptySet();
+                    throw new GradleException(String.format("Failed to resolve the dependencies from %s:%s; " +
+                                    "We can not reliably continue without this information.",
+                            subproject.getPath(), implementationName), e);
```
timtebeek, line 1310: As discussed internally we agreed in principle with throwing an exception; now need to replicate this to the other instances where we call `resolve()`; some of those hadn't yet done any exception handling, so need a bit of time to figure if throwing is even necessary there.

shanman190: So just a question, but is the default Gradle exception not sufficient when resolution fails? In this case, were creating a resolvable configuration where we may have had one that wasn't resolvable.

PR #345, opened 2024-11-14
Simplify Plugins approach to adding dependencies so that it no longer uses a hidden detached configuration.

arlaneenalra:
## What's changed?
Removed dependency copying out of the `rewrite` configuration to a hidden detached one.

## What's your motivation?
We have internal tooling that relies on being able to attach to configurations with a `beforeResolve` hook override dependencies and provide centralized dependency versioning. With the detached configuration there is not way to attach the hook and dependency resolution fails. By removing the detached configuration in favor of the `rewrite` configuration we should be able to use normal hooks as expected.

We've attempted several approaches without modifying the existing plugin but each has run into subtle problems.
* Resolving our internal dependencies first (currently a library of internal recipes) and injecting the resovled files as dependencies. - Lead to duplicated jars on the classpath.
* Replacing the dependencies provider externally with one we can control. - There's no way to get a list of the dependencies the plugin adds without reflectively changing access levels, other forms of nastiness, or repeating the first problem but from the opposite side.

### Checklist
- ~~[ ] I've added unit tests to cover both positive and negative cases~~ _Existing test cases already cover the behavior and the expected behavior should not change._
- [/] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [/] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -73,8 +73,13 @@ public class RewritePlugin implements Plugin<Project> {
 
         // Rewrite module dependencies put here will be available to all rewrite tasks
         Configuration rewriteConf = project.getConfigurations().maybeCreate("rewrite");
-        knownRewriteDependencies(extension, project.getDependencies()).forEach(dep -> {
-            rewriteConf.getDependencies().addLater(project.provider(() -> dep));
+        // Defer actually evaluating the dependencies until resolution is triggered. This should allow
+        // the user to set the rewrite version in the extension. `addLater` doesn't work here because
+        // it operates on a single dependency at a time.
+        rewriteConf.getIncoming().beforeResolve(conf -> {
+            rewriteConf.getDependencies().addAll(
+                knownRewriteDependencies(extension, project.getDependencies())
+            );
         });
 
         // Because of how this Gradle has no criteria with which to select between variants of
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -71,8 +71,38 @@ public class RewritePlugin implements Plugin<Project> {
 
         // Rewrite module dependencies put here will be available to all rewrite tasks
         Configuration rewriteConf = project.getConfigurations().maybeCreate("rewrite");
+        rewriteConf.withDependencies(deps -> {
+            knownRewriteDependencies(extension, project.getDependencies()).forEach(deps::add);
+        });
 
-        Provider<Set<File>> resolvedDependenciesProvider = project.provider(() -> getResolvedDependencies(project, extension, rewriteConf));
+        // Because of how this Gradle has no criteria with which to select between variants of
+        // dependencies which expose differing capabilities. So those must be manually configured
+        try {
+            final ObjectFactory objectFactory = project.getObjects();
+            rewriteConf.attributes(attributes -> {
+                // Adapted from org.gradle.api.plugins.jvm.internal.DefaultJvmEcosystemAttributesDetails
+                attributes.attribute(
+                        Category.CATEGORY_ATTRIBUTE, objectFactory.named(Category.class, Category.LIBRARY));
+                attributes.attribute(Usage.USAGE_ATTRIBUTE, objectFactory.named(Usage.class, Usage.JAVA_RUNTIME));
+                attributes.attribute(
+                        LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
+                        objectFactory.named(LibraryElements.class, LibraryElements.JAR));
+                attributes.attribute(BUNDLING_ATTRIBUTE, objectFactory.named(Bundling.class, Bundling.EXTERNAL));
+                try {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -70,8 +71,38 @@ public void apply(Project project) {
 
         // Rewrite module dependencies put here will be available to all rewrite tasks
         Configuration rewriteConf = project.getConfigurations().maybeCreate("rewrite");
+        rewriteConf.withDependencies(deps -> {
```
shanman190, line 74: According to the documentation, a plugin author should not mutate the dependencies of said configuration within this method.

https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#withDependencies(org.gradle.api.Action)

I think using `conf.getDependencies().addLater(..)` should be workable with the Gradle 4.10 floor from a supported tooling standpoint.

arlaneenalra: Easily done. I'll push an updated version shortly.

arlaneenalra: Done. I also fixed the `dependencyInScript` test that had an incorrect `version` parameter. Jackson has a 2.18.1 version that was causing it to fail. Not sure about the lombok test, other than it appears to be failing on the mainline as well.

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -70,8 +73,38 @@ public void apply(Project project) {
 
         // Rewrite module dependencies put here will be available to all rewrite tasks
         Configuration rewriteConf = project.getConfigurations().maybeCreate("rewrite");
+        rewriteConf.getDependencies().addAllLater(
```
shanman190, line 76: Have you also tried the `testGradle4` task as well? I think `addAllLater` was added in a newer version, so this may break the lower bound of version compatibility.

shanman190: Running the CI task now as well, so we'll know from that as well soon.

arlaneenalra: Didn't know about that task.. and it looks like you're right. Let me fix that.

arlaneenalra: Looks like this is the failing test, which I believe is one of the two that are failing on the mainline. I could fix the other one, but I don't know what's going on with this one:  
```
RewriteRunTest > lombokTypeAttribution(File) FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 
      "package org.openrewrite.before;
  
      import lombok.Getter;
  
      public class N {
          @Getter
          int n;
  
          public static void foo() {
              System.out.println(/*~~>*/new N().getN());
          }
      }"
     but was: 
      "package org.openrewrite.before;
  
      import lombok.Getter;
  
      public class N {
          @Getter
          int n;
  
          public static void foo() {
              System.out.println(new N().getN());
          }
      }"
        at app//org.openrewrite.gradle.RewriteRunTest.lombokTypeAttribution(RewriteRunTest.kt:1564)
 ```

**Comment by arlaneenalra** [comment]
Just realized this is running `knownRewriteDependencies` too soon ... I'll need to fix that.

**Comment by arlaneenalra** [comment]
Ok, switched to `beforeResolve` instead of `addLater` That lombok issue is still present from the mainline. Not sure what to do with that.

**Comment by arlaneenalra** [comment]
Ping, What are the next steps here?

**Comment by shanman190** [comment]
@arlaneenalra, right now it appears that the same failing test on main is causing you some grief here. Just to confirm, it is just the lombok test that fails with the `testGradle4` task? If so, we can probably go ahead and get this merged and fix the test. I think Sam was planning to rework the lombok setup very soon.

**Comment by arlaneenalra** [comment]
> @arlaneenalra, right now it appears that the same failing test on main is causing you some grief here. Just to confirm, it is just the lombok test that fails with the `testGradle4` task? If so, we can probably go ahead and get this merged and afterwards fix the test. I think Sam was planning to rework the lombok setup very soon, so this test would probably get updated due to that.

I'm pretty sure that particular test is actually skipped when using `testGradle4`. Yeah I just verified, it get's skipped for that task. 

```

RewriteRunTest > lombokTypeAttribution(File) SKIPPED
```

Probably because of https://github.com/openrewrite/rewrite-gradle-plugin/blob/95e550b4bdf81e72b3e7822d4d2f71a82a550088/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt#L1503 

It is the only test to fail for the generic `test` and the ci build from what I can tell.

shanman190 (approved):

timtebeek:
hi @arlaneenalra ; thanks again! Your change is now available through
https://github.com/openrewrite/rewrite-gradle-plugin/releases/tag/v6.28.0

Hope that helps you there!

arlaneenalra:
It should, I'm on pto right now, but I should put this to use shory after
the holiday.

Issue #348, opened 2024-11-28
READ_DATE_TIMESTAMPS_AS_NANOSECONDS when rewrite Spring Boot 2.7 to 3.x


nicolasb29:
Hi,

I'm using OpenRewrite to migrate Spring Boot 2.7 to Spring Boot 3.x, like a lot of people I think, but since the version **[6.28.0](https://github.com/openrewrite/rewrite-gradle-plugin/releases/tag/v6.28.0)** I have this error :

`java.lang.RuntimeException: Error while visiting build.gradle: java.lang.NoSuchFieldError: Class com.fasterxml.jackson.annotation.JsonFormat$Feature does not have member field 'com.fasterxml.jackson.annotation.JsonFormat$Feature READ_DATE_TIMESTAMPS_AS_NANOSECONDS'`

I made a project on GitHub to show the problem:

- [URL of the project](https://github.com/nicolasb29/testopenrewritejackson)
- [Output log with the error](https://github.com/nicolasb29/testopenrewritejackson/actions/runs/12074196292/job/33671746482#step:6:2918)
- Rewrite is launch with this command [./gradlew rewriteRun --init-script init.gradle -Drewrite.activeRecipe=org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_0 --info](https://github.com/nicolasb29/testopenrewritejackson/blob/main/.github/workflows/gradle.yml#L41C12-L41C144)

I made a [Gradle dependencies](https://github.com/nicolasb29/testopenrewritejackson/actions/runs/12074196292/job/33671746482#step:5:199) and it seems that `com.fasterxml.jackson` is downgraded from **2.17.2** to **2.13.5** (version in dependency of [Spring Boot 2.7.18](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-dependencies/2.7.18), cause of the problem ?). The annotation `READ_DATE_TIMESTAMPS_AS_NANOSECONDS` was included in version 2.15.

Best regards.

timtebeek:
Thanks for the detailed analysis already @nicolasb29 ; Can't say I've heard of these issues before. Quickly glancing at recipes we seem to have this one: https://github.com/openrewrite/rewrite-migrate-java/blob/cd36280b91d497086e220e965f2af3106e786ebf/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml#L805-L810
[...]

PR #349, opened 2024-11-28
Add jackson-annotations version before resolving

nicolasb29:
https://github.com/openrewrite/rewrite-gradle-plugin/issues/348

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
@@ -221,6 +221,7 @@ public class RewritePlugin implements Plugin<Project> {
                 deps.create("org.openrewrite:rewrite-polyglot:" + extension.getRewritePolyglotVersion()),
                 deps.create("org.openrewrite.gradle.tooling:model:" + extension.getRewriteGradleModelVersion()),
                 deps.create("com.fasterxml.jackson.module:jackson-module-kotlin:" + extension.getJacksonModuleKotlinVersion()),
+                deps.create("com.fasterxml.jackson.core:jackson-annotations:" + extension.getJacksonModuleKotlinVersion()),
                 deps.create("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:" + extension.getJacksonModuleKotlinVersion())
         ).collect(Collectors.toList());
     }


timtebeek:
Thanks @nicolasb29 ! @arlaneenalra Would you care to weigh in here? Since you were involved with 
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/345

See also this problem now seen with "Spring Dependency Management plugin" on older Spring Boot versions
- https://github.com/openrewrite/rewrite-gradle-plugin/issues/348

arlaneenalra:
Sorry for the delay. Based on what I'm reading, these two could be related. I'd be willing to bet the Spring Plugin is doing something similar to what our plugin is doing. You've got me wondering if it would make sense to revert my change and potentially resubmit it behind a flag to disable copying explicitly.

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -190,12 +168,20 @@ private static void configureProject(Project project, RewriteExtension extension
...
+    private Set<File> getResolvedDependencies(Project project, RewriteExtension extension, Configuration rewriteConf) {
+        if (resolvedDependencies == null) {
+            Dependency[] dependencies = Stream.concat(
+                    knownRewriteDependencies(extension, project.getDependencies()).stream(),
+                    rewriteConf.getDependencies().stream()
+            ).toArray(Dependency[]::new);
+            // By using a detached configuration, we separate this dependency resolution from the rest of the project's
+            // configuration. This also means that Gradle has no criteria with which to select between variants of
+            // dependencies which expose differing capabilities. So those must be manually configured
+            Configuration detachedConf = project.getConfigurations().detachedConfiguration(dependencies);
```
arlaneenalra, line 180: Would it be possible to add a flag to the extension to disable the detached configurations behavior?

shanman190: Potentially? The unfortunate part of doing that would mean moving further away from being compatible with the configuration cache feature (all tasks are marked as not compatible presently).

nicolasb29: > Would it be possible to add a flag to the extension to disable the detached configurations behavior?
> 
> Or detect the spring plugin and use only use the detached configuration if it's present?

I will check

nicolasb29: I add a Gradle parameter so you can add `-Prewrite.detachResolve` to detach the resolver.

shanman190: @arlaneenalra, I'm trying to wrap my head around this, but is there a reason for your all's plugin to try to bother with central dependency version management? It seems like the `rewrite` configuration not being a detached configuration in all cases is causing some pretty wide spread problems.

The `rewrite` configuration is also really meant to be isolated just for the purposes of this plugin, so other plugins manipulating it's versions has shown to be detrimental (case in point from @nicolasb29's issue).

I'm really wondering if just going back to always being a detached configuration may be the right thing to do rather than trying to have a toggle to go one way or the other...

arlaneenalra: >@arlaneenalra, I'm trying to wrap my head around this, but is there a reason for your all's plugin to try to bother with central dependency version management for the rewrite configuration specifically? It seems like the rewrite configuration not being a detached configuration in all cases is causing some pretty wide spread problems.

For our use case, we have a local publishing process that redirects some dependencies when doing development. That process doesn't work for an detached configuration like this without special handling (i.e. a custom configuration class). Our work arounds wound up injecting conflicting dependency jars and causing random failures because of the conflicts.  If this is really not viable, I'll have to do some more digging to see if there are other alternatives .. hmm

shanman190: For what it's worth, what I typically do is use Gradle's `publishToMavenLocal`, then bring in the local snapshot jar via Gradle's native dependency resolution.

timtebeek:
Thanks all! Given the volume of issues reported with the earlier change we've decided to rollback instead:
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/353

I hope that gets most of the folks going again; apologies for the disruption here.

PR #355, opened 2024-12-15
Apply the MSAL license

jkschneider:
See [the OpenRewrite licensing doc](https://docs.openrewrite.org/licensing/openrewrite-licensing) for an explanation of why this is being done and what it means.


jbduncan:
Oh that's awfully gutting, but I understood the business justification. 🥲👌

shanman190:
@jbduncan, FYI, this has effectively been reverted as you can see here:
https://github.com/openrewrite/rewrite-gradle-plugin/pull/364

PR #369, opened 2025-02-10
First stab at file inclusions

Laurens-W:
## What's changed?

Added inclusion as a counter to exclusions

@sambsnyd 

### Checklist


sambsnyd:
So this is good but does it have exactly the same semantics as the `FileInclusion` we use in the CLI? If not we're inviting bugs and drift. Can we move that class into rewrite-polyglot and reuse it here, in the maven plugin, and in the Gradle plugin?

PR #381, opened 2025-05-15
Add test case for gradle.properties

sullis:
## What's changed?

new test case


timtebeek:
Thanks! Were we missing coverage here? We tend to test recipes closer to their implementation where possible, unless there's a gradle specific regression for reason there that I'm not aware of. :)

sullis:
> Thanks! Were we missing coverage here? We tend to test recipes closer to their implementation where possible, unless there's a gradle specific regression for reason there that I'm not aware of. :)

I am moving this PR to Draft status.  I am investigating an issue in my local environment. 

I'll share an update when I have a better understanding.

PR #393, opened 2025-08-01
refactor: Update Gradle wrapper

shanman190:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.gradle.UpdateGradleWrapper?organizationId=QUxML09wZW4gU291cmNlL09wZW5SZXdyaXRl#defaults%3DW3sidmFsdWUiOmZhbHNlLCJuYW1lIjoiYWRkSWZNaXNzaW5nIn1d


shanman190:
Most of the changes were related to tests. There was a need to be able to perform a range (min-max), so I went ahead and introduced the `org.openrewrite.gradle.condition.*` classes -- mirroring `@DisabledForJreRange`/`@EnabledForJreRange` as a source for the example. We could optionally move these to `openrewrite/rewrite-gradle` for wider availablility.

I did slightly alter some of the test cases such that we could retain a single test that covered the range, but in some places introducing a new test for the newer Gradle version functionality was unavoidable.

sambsnyd (approved):

PR #400, opened 2025-08-18
Add `.gitignore` support for exclusions

shanman190:
## What's changed?
Add support for `.gitignore` exclusions

Nothing in particular

Anybody

None

### Checklist


plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
```diff
@@ -1335,7 +1355,33 @@ private SourceFile logParseErrors(SourceFile source) {
...
+                            return true;
+                        } else {
+                            walk.enterSubtree();
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+        }
```
timtebeek, line 1384: Would it make sense to do the PathMatcher checks first? I imagine those are cheaper to do, and quite likely to match if configured.

PR #420, opened 2025-11-26
Make sure that user projects can resolve dependency variants correctly

wjglerum:
## What's changed?
Enables the `id("jvm-ecosystem")` plugin in all projects to make sure Gradle can resolve the correct variant of a dependency.

## What's your motivation?
- This should resolve troubles in user projects see: 
  - Closes https://github.com/openrewrite/rewrite-gradle-plugin/issues/421 
  - Closes https://github.com/openrewrite/rewrite-github-actions/issues/161#issuecomment-3550632631

Is there a test case/project where we could try this out?

@timtebeek 

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,6 @@ jobs:
         run: ./gradlew ${{ env.GRADLE_SWITCHES }} build
       - name: publish-snapshots
         if: github.event_name != 'pull_request'
-        uses: gradle/actions/setup-gradle@v5
-      - name: snapshot
         run: ./gradlew ${{ env.GRADLE_SWITCHES }} snapshot :plugin:publishPluginMavenPublicationToSonatypeRepository -x test -x publishPlugins
         env:
           ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.SONATYPE_USERNAME }}


wjglerum:
The failure is for the publishing of the snapshot now see https://github.com/openrewrite/rewrite-gradle-plugin/actions/runs/19700227861/job/56434541510#step:7:368
Is that expected @timtebeek ?

timtebeek:
Thanks for the help here! I've tagged @shanman190 for review as he's more in tune with any edge cases we might encounter here. As a quick thought we did add support for Android a long time ago, but I don't think that's actively used, so perhaps it can't hurt to add this for now at least to unblock folks affected by the linked issues.

plugin/src/main/java/org/openrewrite/gradle/RewritePlugin.java
```diff
@@ -98,6 +99,11 @@ public void apply(Project project) {
     private static void configureProject(Project project, RewriteExtension extension, TaskProvider<RewriteDryRunTask> rewriteDryRun, TaskProvider<RewriteRunTask> rewriteRun) {
         // DomainObjectCollection.all() accepts a function to be applied to both existing and subsequently added members of the collection
         // Do not replace all() with any form of collection iteration which does not share this important property
+        try {
+            project.getPluginManager().apply(JvmEcosystemPlugin.class);
+        } catch (NoClassDefFoundError e) {
+            // Gradle version is too old to have the JvmEcosystemPlugin, that's OK
+        }
```
shanman190: I probably would have done a `GradleVersion` comparison, but this works.

shanman190 (approved):
This'll add the attribute `org.gradle.jvm.environment=standard-jvm` as desired (also some other various attributes), but that won't cause any other edge case issues.

Issue #382, opened 2025-05-25
Log Recipe names if recipe validation error occirs


danish-kamal-sophos:
## What problem are you trying to solve?
Log Recipe names if recipe validation error occirs

## Describe the solution you'd like
Recipe validation errors that are logged upon recipe activation:

```
Recipe validation error in newGroupId: is required
Recipe validation error in newArtifactId: is required
Recipe validation error in coordinates: newGroupId OR newArtifactId must be different from before
Recipe validation error in newGroupId: is required
Recipe validation error in newArtifactId: is required
Recipe validation error in coordinates: newGroupId OR newArtifactId must be different from before
Recipe validation error in newGroupId: is required
Recipe validation error in newArtifactId: is required
Recipe validation error in coordinates: newGroupId OR newArtifactId must be different from before
Recipe validation error in newGroupId: is required
Recipe validation error in newArtifactId: is required
Recipe validation error in coordinates: newGroupId OR newArtifactId must be different from before
Recipe validation errors detected as part of one or more activeRecipe(s). Execution will continue regardless.
```

do not contain the recipe name. It would be nice to log recipe name to make debugging and fix easier.

## Have you considered any alternatives or workarounds?
Tried debug/info logging but seems there is no additional logging.

## Additional context

Message is logged from DefaultProjectParseer.java here: https://github.com/openrewrite/rewrite-gradle-plugin/blob/main/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java#L1381

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes!

PR #423, opened 2025-12-14
Improve error logging for recipe validation failures by including recipe name

RaoPrashanth:
## What's changed?
Improved error logging for recipe validation failures by including the recipe name in the error message. When recipe validation fails, the error log now shows which specific recipe caused the validation error, making it much easier to debug configuration issues.

**Before:**
```
Recipe validation error in {property}: {message}
```

**After:**
```
Recipe validation error in {recipeName} for property {property}: {message}
```

## What's your motivation?
When recipe validation errors occur, the previous error messages didn't indicate which recipe was causing the problem. This made debugging difficult, especially in builds with multiple active recipes. By adding the recipe name to the error message, developers can immediately identify the problematic recipe without having to guess or step through code.

This improvement enhances the developer experience by providing clearer, more actionable error messages.

- closes https://github.com/openrewrite/rewrite-gradle-plugin/issues/382

- The error message format - does it provide enough context?
- Whether removing the exception parameter from the log statement is appropriate (it wasn't being used in the format string anyway)

N/A

Alternative considered: Including the full recipe descriptor with all options, but this would make the error message too verbose. The recipe name provides the essential context needed for debugging while keeping the message concise.

## Any additional context
This is a small quality-of-life improvement that doesn't change any functionality - it only enhances the error reporting to make troubleshooting easier.

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases (N/A - logging change only)
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices) (N/A - not a recipe change)

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java
@@ -1435,7 +1435,12 @@ public class DefaultProjectParser implements GradleProjectParser {
         List<Validated.Invalid<Object>> failedValidations = validated.stream().map(Validated::failures)
                 .flatMap(Collection::stream).collect(toList());
         if (!failedValidations.isEmpty()) {
-            failedValidations.forEach(failedValidation -> logger.error("Recipe validation error in {} for property {}: {}", recipe.getName(), failedValidation.getProperty(), failedValidation.getMessage()));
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Thanks for picking this up @RaoPrashanth ! Should help others troubleshoot more effectively

PR #432, opened 2026-03-30
Fix isExcluded gitignore handling using shared GitIgnore utility

pdelagrave:
Fixes two bugs in `isExcluded`:

1. A recursive call prepending `/` for PathMatcher compatibility broke the JGit TreeWalk path comparison (JGit paths never have a leading slash), making the gitignore check dead code.
2. Only `isEntryIgnored()` was checked without consulting the git index, so tracked files matching `.gitignore` would have been incorrectly excluded if the path bug were fixed.

Replaces the inline JGit implementation with the shared `GitIgnore.isIgnoredAndUntracked()` utility from rewrite-core (openrewrite/rewrite#7215).

Depends on openrewrite/rewrite#7215.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -91,6 +91,9 @@ val rewriteDependencies = configurations.create("rewriteDependencies")
 configurations.named("compileOnly").configure {
     extendsFrom(rewriteDependencies)
 }
+configurations.named("testRuntimeOnly").configure {
+    extendsFrom(rewriteDependencies)
+}
 
 dependencies {
     "rewriteDependencies"(platform("org.openrewrite:rewrite-bom:$latest"))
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/openrewrite/gradle/isolated/GitExclusions.java
```diff
@@ -0,0 +1,97 @@
...
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.util.Collection;
+
+import static org.openrewrite.PathUtils.separatorsToUnix;
+
+class GitExclusions {
+
+    static boolean isExcluded(@Nullable Repository repository, Collection<PathMatcher> exclusions, Path path) {
```
shanman190: We may want to inline this. There's not really much utility to having it be a new class even if package private.

pdelagrave: I didn't review claude's work yet hence the draft state but I agree.
Claude did it for testability, `DefaultProjectParser` wasn't testable as is due to some of its dependencies not being on the test classpath...

Anyway the bulk of the code will be in rewrite so it's reusable by the maven plugin who has the same issue.

PR #433, opened 2026-04-11
Exclude generated sources in build directory from parsing

### LINKED ISSUE #216: Generated kotlin sources inside the build directory are showing up in recipe results
Sources inside a build directory, usually produced by a codegen process like protobuf, are being included in recipe results. Anything inside a build folder should be treated as a generated file and not have recipes executed on it

timtebeek:
- Fixes #216.

## Summary
- Source directories inside the build directory (e.g. protobuf-generated Kotlin sources) are now filtered out at the directory level before file walking, preventing generated sources from being parsed and appearing in recipe results
- Type information from generated sources remains available via compiled class files already on the classpath
- Removed the hardcoded `/protos/build/generated` exclusion, now covered by the generic build directory filter
- Consolidated duplicated `isExcluded`/`buildDir` filtering from individual parse methods (`parseJavaFiles`, `parseKotlinFiles`, inline Groovy/multiplatform blocks) into a single filter at the stream assembly point, in both `DefaultProjectParser` and `AndroidProjectParser`

## Test plan
- [x] Existing tests pass (`./gradlew :plugin:test`)

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/isolated/AndroidProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/isolated/AndroidProjectParser.java
@@ -136,8 +136,6 @@ class AndroidProjectParser {
                     alreadyParsed.addAll(javaPaths);
                     Stream<SourceFile> parsedJavaFiles = parseJavaFiles(javaPaths,
                             ctx,
-                            buildDir,
-                            exclusions,
                             javaSourceCharset,
                             javaVersion,
                             dependencyPaths,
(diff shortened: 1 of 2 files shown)


shanman190:
Just curious, but should isExcluded be extracted from each individual parse chain and moved toward the end of all parsing to eject any file that happens to be explicitly excluded?

timtebeek:
Good suggestion! Extracted the `isExcluded`/`buildDir` filtering from each individual parse method (`parseJavaFiles()`, `parseKotlinFiles()`, inline Groovy block, and multiplatform Kotlin block) into a single filter applied at the stream assembly point. This applies to both `DefaultProjectParser` and `AndroidProjectParser`.

shanman190 (approved):

Issue #388, opened 2025-07-04
Modernize the Gradle Plugin


shanman190:
## What problem are you trying to solve?

We receive issues from time to time about not supporting one or another Gradle feature properly, using the entire Gradle build as if it were a single project, not supporting configuration cache, inputs and outputs are missing in places, etc. All of this comes to a point when we're attempting to support such a large version range and as Gradle major versions introduce new defaults.

## Describe the solution you'd like

We should really rewrite the plugin from the ground up such that it can work over a single project at a time, has all inputs and outputs modeled correctly, takes advantage of lazy task registration and property configuration, is compatible with configuration cache, etc.

## Have you considered any alternatives or workarounds?

At some point we're going to run out of workarounds, so this is trying to stay ahead of that time.

## Additional context
N/A

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
Always

timtebeek:
we have more freedom to restructure the plugin now that the CLI no longer depends on it. We might even choose to drop support for the oldest of Gradle versions.

PR #439, opened 2026-04-12
Declare classpath as @Classpath input using ConfigurableFileCollection

### LINKED ISSUE #387: Dependencies resolved too early - breaks gradlex-org/extra-java-module-info plugin
I am using this plugin together with [gradlex-org/extra-java-module-info](gradlex-org/extra-java-module-info/).

When running, I get

```
caused by: java.lang.RuntimeException: File does not exist: /home/runner/work/jabref/jabref/test-support/build/libs/test-support.jar
  This is likely because a tool or another plugin performs early dependency resolution.
  You can prevent this error by setting 'skipLocalJars = true'.
```

https://github.com/JabRef/jabref/actions/runs/16052632080/job/45298938588

The workaround is to execute all GUI tests in the before openrewrite:

    xvfb-run --auto-servernum ./gradlew --no-configuration-cache :jabgui:test :rewriteDryRun

Instead of

    ./gradlew --no-configuration-cache :rewriteDryRun

(See https://github.com/JabRef/jabref/pull/13448)

I filed an issue at that plugin (https://github.com/gradlex-org/extra-java-module-info/issues/188), but the response is that I should ask here:

> If I read the code correctly the plugin is [doing dependency resolution ](https://github.com/openrewrite/rewrite-gradle-plugin/blob/main/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java?rgh-link-date=2025-07-04T06%3A13%3A31.000Z#L82)(accessing files on a classpath), [without declaring the that classpath input to the task](https://github.com/openrewrite/rewrite-gradle-plugin/blob/main/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java?rgh-link-date=2025-07-04T06%3A13%3A31.000Z#L73).

Thus, the question if you could add `classpath` as required input.

timtebeek:
## Summary
- Fixes #387.

The rewrite tasks accessed classpaths without declaring them as Gradle task inputs, so Gradle could not infer task dependencies. This caused "File does not exist" errors when other plugins (e.g. `extra-java-module-info`) produced jars that needed to be built first.

- Replace `Provider<Set<File>> resolvedDependencies` with a `ConfigurableFileCollection` annotated with `@Classpath`, wired via `.from(detachedConf)` using a `Callable` for lazy creation. This gives Gradle full provenance of the rewrite recipe jars.
- Add a separate `projectClasspath` `@Classpath` input, wired from each source set's compile and runtime classpaths. This lets Gradle infer that dependent project jars (e.g. `test-support.jar`) must be built before rewrite tasks run.

## Test plan
- [x] `./gradlew :plugin:test` — all tests pass
- [x] End-to-end validation on three projects: single-module, multi-module, and custom `rewrite` configuration dependency

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/AbstractRewriteTask.java
@@ -16,8 +16,9 @@
 package org.openrewrite.gradle;
 
 import org.gradle.api.DefaultTask;
+import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.ProjectLayout;
-import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.Classpath;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.options.Option;
@@ -25,16 +26,13 @@ import org.gradle.util.GradleVersion;
 import org.jspecify.annotations.Nullable;
 
 import javax.inject.Inject;
-import java.io.File;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Set;
 
-import static java.util.Collections.emptySet;
 import static java.util.stream.Collectors.toSet;
 
 public abstract class AbstractRewriteTask extends DefaultTask {
-    protected @Nullable Provider<Set<File>> resolvedDependencies;
     protected boolean dumpGcActivity;
     protected @Nullable GradleProjectParser gpp;
     protected @Nullable RewriteExtension extension;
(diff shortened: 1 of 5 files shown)


shanman190:
There's a much deeper problem here in that we implicitly depend on numerous pieces of information on the project (ex: all configurations to generate the `GradleProject` marker).

So we'll not be properly modeling those and still be missing relationships. This is what led me to create an overarching [issue](https://github.com/openrewrite/rewrite-gradle-plugin/issues/388) to address these and many more aspects that we haven't yet been able to address.

timtebeek:
Thanks yes was aware of that issue, but wasn't sure if we could incrementally work our way there with steps like these; you'll know best if we could, so let me know if this should be refined or closed.

timtebeek:
Closing as it's not clear what the next step would be; that said we have more freedom to restructure the plugin now that the CLI no longer depends on it. We might even choose to drop support for the oldest of Gradle versions.

PR #440, opened 2026-04-17
Add testGradle9 task pinned to Gradle 9.4.1 on Java 25

timtebeek:
## Summary
- Adds a `testGradle9` task modeled on the existing `testGradle4` / `testGradle8` tasks, pinned to Gradle `9.4.1` (the current stable release; 9.5 is still in RC).
- Runs the suite on a Java 25 toolchain, since Gradle 9.1+ supports Java 25 at runtime.
- Wires the task into `check`, so CI (which runs `./gradlew build`) picks it up automatically — no workflow change needed.

## Why
The default `test` task runs against whatever Gradle version the wrapper ships (currently 9.3.1). The dedicated `testGradleN` tasks exist to pin coverage against specific Gradle versions so wrapper upgrades don't silently change our support surface. With Gradle 9.5.0-rc-3 out and the wrapper about to move forward along 9.x, pinning 9.4.1 gives us a deliberate "latest stable 9.x" target, mirroring how `testGradle8` pins the latest 8.x.

## Test plan
- [x] `./gradlew :plugin:tasks --all` shows `testGradle9` registered.
- [x] `./gradlew :plugin:check --dry-run` shows `testGradle4`, `testGradle8`, and `testGradle9` scheduled.
- [ ] CI runs `testGradle9` as part of `build` on Java 25 and passes.

### DIFF
--- a/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
+++ b/plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
@@ -1041,6 +1041,9 @@ class RewriteRunTest : RewritePluginTest {
                        url = uri("https://central.sonatype.com/repository/maven-snapshots")
                     }
                 }
+                kotlin {
+                    jvmToolchain(21)
+                }
                 rewrite {
                     activeRecipe("org.openrewrite.test.FindString")
                 }
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -215,13 +215,12 @@ val testGradle8 = tasks.register<Test>("testGradle8") {
     })
 }
 val testGradle9 = tasks.register<Test>("testGradle9") {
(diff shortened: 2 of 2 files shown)


plugin/src/test/kotlin/org/openrewrite/gradle/RewriteRunTest.kt
```diff
@@ -1041,6 +1041,9 @@ class RewriteRunTest : RewritePluginTest {
                        url = uri("https://central.sonatype.com/repository/maven-snapshots")
                     }
                 }
+                kotlin {
+                    jvmToolchain(21)
```
timtebeek, line 1045: I'm surprised we're only discovering this failure now; Especially since we'd used `@EnabledForGradleRange(min = "9.0")` above.

greg-at-moderne (approved):

PR #456, opened 2026-07-26
Reload recipe classes when a classpath entry changed

### LINKED ISSUE #453: Java recipes are not getting updated

I am using

- OpenRewrite v3.34.0 (BOM)
- Gradle plugin v7.36.0
- rewrite-java v`[whatever version the BOM is using]`

I am using the Gradle plugin, and my project is a single module project with an included build.

A repro can be found here: https://github.com/freya022/rewrite-gradle-plugin/tree/repro/retained-rewrite-classpath

1. Run `./gradlew :rewriteDryRun`, keep the logs
2. Modify the prints in `recipes/src/main/kotlin/org/example/TestRecipe.kt`, `println("Test 1")` => `println("Test 2")`
3. Run the same command again, notice the logs are the same, the code was not updated

The recipe should be updated

## What did you see instead?

The recipe was not updated and kept running old versions until Gradle was fully restarted

## What is the full stack trace of any errors you encountered?

N/A

I'm fairly sure this is an issue of the `ClassLoader` not being replaced appropriately (adding a dependency in the `rewrite` configuration updates the recipe), but I'm not sure what the actual fix would be.

From
https://github.com/openrewrite/rewrite-gradle-plugin/blob/b3a0c6443e1e15d7556d27208c27a9dbc3b39258/plugin/src/main/java/org/openrewrite/gradle/DelegatingProjectParser.java#L65-L73
I can see some potential issues, if we assume the plugin class loader doesn't change when only my recipe changes, the classpath URLs will remain the same (only the contents changed) and thus it will reuse the old loader.

timtebeek:
- Fixes #453

### The problem

`DelegatingProjectParser` caches the `RewriteClassLoader` in a static field, so it survives across builds in the Gradle daemon, and only replaced it when the *set of classpath locations* changed:

```java
if (rewriteClassLoader == null ||
        !classpathUrls.equals(rewriteClasspath) ||
        rewriteClassLoader.getPluginClassLoader() != pluginClassLoader) {
```

Recipes built by the project itself (an included build, a `project(":recipes")` dependency, or `publishToMavenLocal`) keep the same location when rebuilt, so the URLs compare equal and the previous recipe classes keep being used until the daemon is stopped. @freya022's diagnosis in the issue was spot on.

### The fix

Compare the size and last modified time of each classpath entry as well; for directory entries that's rolled up over the files they contain. Two `stat` calls per classpath entry is nothing next to parsing a project, and published artifacts in the module cache never change, so ordinary users see no extra class loading.

### Metaspace

Replacing the class loader more often surfaced that replaced loaders are never unloaded. One cause is in this repo: JGit starts a `JGit-WorkQueue` daemon thread per class loader that initializes it, and the thread's pool references the loader that created it. Without the shutdown added here, thread count grows one per replacement and every ~33 MB loader is pinned forever; with Gradle's default 384 MB metaspace the daemon dies after a handful of recipe edits.

Shutting the work queue down keeps the thread count at 1, but measuring across 8 edit/run cycles the loaders are still retained (~40 MB per cycle, `VM.classloader_stats` shows 8 live `RewriteClassLoader`s that survive a forced full GC). So at least one more retainer lives outside this repo — most likely a `ThreadLocal` or static registered from rewrite on a daemon thread. That retention is pre-existing: any change to the `rewrite` configuration already replaced the loader. This PR makes it easier to hit while developing recipes, so it is worth a follow-up on the rewrite side. Until then, recipe authors seeing `Metaspace` errors can raise `org.gradle.jvmargs=-XX:MaxMetaspaceSize=1g`.

### Verification

- `DelegatingProjectParserTest` covers the fingerprint: unchanged classpath, order independence, replaced jar, same-size rebuild, and changed/added files inside a directory entry.
- `RewriteDiscoverTest` builds a project with a recipe subproject, runs `rewriteDiscover`, swaps the recipe source, and runs again in the same TestKit daemon. It fails on `main` and passes here.
- Ran @freya022's scenario (included build + dependency substitution) against a locally published plugin: before, `rewriteRun` kept applying the first version of the recipe across daemon-reusing builds; after, every edit takes effect.

### DIFF
--- a/plugin/src/main/java/org/openrewrite/gradle/DelegatingProjectParser.java
+++ b/plugin/src/main/java/org/openrewrite/gradle/DelegatingProjectParser.java
@@ -19,24 +19,33 @@ import org.gradle.api.Project;
 import org.gradle.internal.service.ServiceRegistry;
 import org.jspecify.annotations.Nullable;
 
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.toList;
 
 public class DelegatingProjectParser implements GradleProjectParser {
+    private static final AtomicLong UNREADABLE_CLASSPATH_ENTRIES = new AtomicLong();
+
     @Nullable
-    protected static List<URL> rewriteClasspath;
+    protected static List<String> rewriteClasspathFingerprint;
     @Nullable
     protected static RewriteClassLoader rewriteClassLoader;
     protected final GradleProjectParser gpp;
@@ -61,15 +70,16 @@ public class DelegatingProjectParser implements GradleProjectParser {
             classpathUrls.add(currentJar);
 
             ClassLoader pluginClassLoader = getPluginClassLoader(project);
+            List<String> classpathFingerprint = fingerprint(classpathUrls);
 
             if (rewriteClassLoader == null ||
-                    !classpathUrls.equals(rewriteClasspath) ||
+                    !classpathFingerprint.equals(rewriteClasspathFingerprint) ||
                     rewriteClassLoader.getPluginClassLoader() != pluginClassLoader) {
                 if (rewriteClassLoader != null) {
-                    rewriteClassLoader.close();
+                    discard(rewriteClassLoader);
                 }
                 rewriteClassLoader = new RewriteClassLoader(classpathUrls, pluginClassLoader);
-                rewriteClasspath = classpathUrls;
+                rewriteClasspathFingerprint = classpathFingerprint;
             }
 
             Class<?> gppClass = Class.forName("org.openrewrite.gradle.isolated.DefaultProjectParser", true, rewriteClassLoader);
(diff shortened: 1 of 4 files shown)


freya022:
Oops, made #457 then saw this in my notifications

timtebeek:
Thanks for pitching in! Do let me know if you have any thoughts on the implementation here as compared to #457!

shanman190:
Just a reminder, but the last time we tried this we had issues with the classloader getting stuck due to classes that were cached in unreachable places. This is where the advisement of not using the daemon originated from.

timtebeek:
Good reminder, and the history backs it up: b89ca68 and 8230068 both tried to close the loader and were reverted the next day, and #445 bumped the TestKit fixture to 1g metaspace in May for the same reason — recipe classes loaded into a daemon aren't unloaded between builds.

What I measured on this branch, for the record:

- ~33 MB per `RewriteClassLoader`; `VM.classloader_stats` showed 8 live after 8 edit/run cycles, surviving two forced full GCs.
- JGit starts a `JGit-WorkQueue` thread per loader whose pool references the loader that created it — one retainer per reload, addressed here; thread count now stays at 1.
- The remaining retention reproduces with `rewriteDiscover` alone (4 cycles, 4 retained loaders) and leaks no threads, so it isn't the parse path.

timtebeek:
Hmm; thanks both, that makes me err on the side of caution and hold off on supporting this particular use case for now. Maybe we can revisit in the future, but for now the risk isn't worth the usability improvement for same-project recipes under development.

freya022:
If we can't fix it for now, what's the recommended way to make Java recipes? Are these issues documented anywhere? ["Local publishing for testing"](https://docs.openrewrite.org/authoring-recipes/recipe-development-environment#local-publishing-for-testing) says to publish the recipe in a local repository, which suffers from the same issue, unless the artifact's version changes every time you do a change (it seems to be implied in "Running your Recipes" when using the recipe starter, but it's not very clear where the `0.1.0-dev.25.uncommitted+f58d7fa` example comes from).

freya022:
I tested your PR and shutting down the JGit pool does remove the reference, though maybe it could be worth running `NLS.clear();` too. (that's what `Git#shutdown` does)

I've checked with VisualVM and `jcmd`, somehow the old classes still doesn't get unloaded, even though it says the `RewriteClassLoader` doesn't retain anything, and no GC root points to it... (from my testing it seems to unload the classes after a while, likely when it's under pressure, not sure if there is a time component to it, sometimes it unloads after the 2nd use, sometimes after the 4th)

That said, I don't get why your PR can't go through (what's the risk?), it won't change anything for people who have their recipes already made (it'll keep reusing the same loader, same as before), and for those developing them, all it does is enable them to reuse their Gradle instances until it runs out of metaspace, instead of *having* to use `--no-daemon` for every try. To me, this just seems like a net win.

If your tests require running without a daemon (literally a daemon that's shutdown after execution), then so be it, but other developers would *really* benefit from this.

Edit: The class loader does get fully unloaded **consistently** on GC calls, but only if you use Shenandoah GC, all other GCs do report them as unreachable but don't reclaim their memory until some point.

plugin/src/main/java/org/openrewrite/gradle/DelegatingProjectParser.java
```diff
@@ -134,6 +146,61 @@ public void shutdownRewrite() {
         });
     }
 
+    private static void discard(RewriteClassLoader classLoader) throws IOException {
+        try {
+            Class.forName("org.openrewrite.gradle.isolated.DefaultProjectParser", true, classLoader)
+                    .getMethod("shutdownJGitWorkQueue")
+                    .invoke(null);
+        } catch (ReflectiveOperationException | LinkageError ignored) {
```
freya022, line 154: ~~`Method#invoke` always throws an `InvocationTargetException`, and I believe the wrapped exception would be a `NoClassDefFoundError` when JGit is absent (though if it was, `DefaultProjectParser` would entirely fail to load, no?)~~

Edit: Nevermind, somehow missed that `InvocationTargetException` inherits from `ReflectiveOperationException`

PR #457, opened 2026-07-26
Compare against a snapshot of the rewrite classpath's content before recreating a new `RewriteClassLoader`

freya022:
## What's changed?
This makes a snapshot of the `rewrite` configuration and checks if the contents changed on subsequent runs, meaning the class loader gets updated every time the recipes do.

## What's your motivation?
- Fixing #453

- I've tested it against the issue, is there really a need to make unit tests for this?
- Should this support snapshotting folders? Currently it assumes only JARs will be present as I've only seen those, if folders are present, it will behave the same as before.

Running a cold Gradle VM everytime when developing recipes.

### Checklist
- [X] I've used the IntelliJ IDEA auto-formatter on affected files


freya022:
Closing in favor of #456

PR #464, opened 2026-08-10
Bump Kotlin plugin to 2.4.0 to match Gradle 9.7.0

timtebeek:
CI on `main` broke after the Gradle wrapper update to 9.7.0: `:plugin:compileKotlin` fails because Gradle 9.7.0 ships `kotlin-stdlib`/`kotlin-reflect` 2.4.0 on the `gradleApi()` classpath, and the pinned Kotlin 2.2.0 compiler can only read metadata up to 2.3.0.

```
e: Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin.
   The actual metadata version is 2.4.0, but the compiler version 2.2.0 can read versions up to 2.3.0.
```

Bumping the Kotlin JVM plugin to 2.4.0 lines the compiler up with what Gradle embeds. Verified locally on Gradle 9.7.0: `:plugin:compileKotlin`, `build` (minus the test-kit suites) and the test compilation all succeed.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -5,7 +5,7 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.util.*
 
 plugins {
-    id("org.jetbrains.kotlin.jvm") version "2.2.0"
+    id("org.jetbrains.kotlin.jvm") version "2.4.0"
     id("com.gradle.plugin-publish") version "latest.release"
     id("com.github.hierynomus.license") version "0.16.1"
     id("nebula.maven-apache-license")


timtebeek:
### Why bumping the Kotlin plugin to 2.4.0 is safe for `rewrite-kotlin` (built against Kotlin 2.3.x)

**Recipe execution is insulated from this bump.** Recipes run inside `RewriteClassLoader`, which is *child-first*: it only delegates to the parent classloader for a short allowlist — `org.openrewrite.gradle.GradleProjectParser`, `RewriteExtension`, `DefaultRewriteExtension`, `org.slf4j`, `org.gradle`, `groovy`, `org.codehaus.groovy`. `kotlin.*` is not on that list, so it resolves from the recipe classpath URLs, and `DefaultProjectParser` (hence `KotlinParser`) is asserted to load from that loader.

The recipe classpath resolves to:

```
org.openrewrite:rewrite-kotlin:8.89.0-SNAPSHOT
 +--- org.jetbrains.kotlin:kotlin-compiler-embeddable:2.3.20
 +--- org.jetbrains.kotlin:kotlin-reflect:2.3.20
 \--- org.jetbrains.kotlin:kotlin-stdlib:2.3.20
```

Those 2.3.20 artifacts are what `rewrite-kotlin` runs against, regardless of which compiler built the plugin. The plugin's own Kotlin is a single main-source file (`GradleProjectSpec.kt`) plus tests.

**What actually changes for consumers.** The plugin's declared runtime dependency on `org.jetbrains.kotlin:kotlin-stdlib` moves 2.2.0 → 2.4.0 (diffed via `:plugin:dependencies --configuration runtimeClasspath` across the commit). That lands on the *buildscript* classpath, not the recipe classpath, and is benign because:

- The `incompatible version of Kotlin` metadata error is a **compile-time** constraint only — a ≥2.4 compiler is needed to compile *against* 2.4.0-metadata classes. Running older-compiled bytecode on a newer stdlib is fine; the stdlib is binary-backward-compatible.
- The plugin already shipped a stdlib newer than Gradle's embedded one (2.2.0 vs Gradle 8.14's 2.0.x) without breaking Kotlin DSL script compilation, so the classpath-ordering concern is already settled empirically.
- CI here exercised Gradle 4.10, 8.14.3 and 9.4.1 test-kit suites in addition to Gradle 9.7.0 itself, all green.

**Worth remembering:** the coupling now runs the other way — while on Gradle 9.7 the plugin's compiler is pinned *forward* by Gradle's embedded Kotlin, so we can't drop back to 2.2/2.3 to match `rewrite-kotlin` without `-Xskip-metadata-version-check`. And if `kotlin.*` were ever added to `PARENT_LOADED_PACKAGES`, the 2.4.0-vs-2.3.20 split would start to matter. Today it doesn't.

PR #467, opened 2026-08-19
Cut over to the Code Genome Project for publishing and org.openrewrite resolution

timtebeek:
Completes the cutover to the Code Genome Project: snapshots publish there instead of Sonatype, nothing publishes to Sonatype or Maven Central at all, and `org.openrewrite`/`io.moderne` artifacts are no longer consumed from Maven Central.

## Publishing

- `plugin/build.gradle.kts` registers CGP's S3 bucket as a Maven publishing repository, mirroring [`org.openrewrite.build.publish-cgp`](https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/main/src/main/java/org/openrewrite/gradle/RewriteCgpPublishPlugin.java) — this repo can't apply that plugin, since it wires nebula and `gradle-nexus.publish-plugin` directly. The host is region-qualified because Gradle's S3 transport otherwise defaults to `us-east-1`.
- `.github/workflows/ci.yml` publishes `:plugin:publishPluginMavenPublicationToCgpRepository` with the `CGP_AWS_*` org secrets, replacing the Sonatype and OSSRH signing credentials. The signing properties were dead already — no `signing` plugin is applied anywhere in this build.
- `build.gradle.kts` drops `io.github.gradle-nexus.publish-plugin` and its `nexusPublishing` block. **These were vestigial:** `org.openrewrite:plugin` was never on Maven Central (`repo1.maven.org` 404s for it), because releases go to the Gradle Plugin Portal via `publishPlugins` and nothing ever invoked `publishToSonatype`. `cgp` is now the only Maven publishing repository.

## Resolution

- Removes the `central.sonatype.com/repository/maven-snapshots` repository.
- Excludes `org.openrewrite` and `io.moderne` from both `mavenCentral` and `gradlePluginPortal` (which proxies Central) whenever CGP credentials are configured, so those groups resolve from CGP alone. This follows the `codegenomeConfigured` pattern from [`rewrite-build-gradle-plugin`](https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/main/build.gradle.kts) — gating on credentials keeps fork PRs, which receive no secrets, building against Central.
- `mavenCentral` stays for third-party dependencies.

## Verification

Run locally with real CGP credentials, so the credentialed path was genuinely exercised rather than falling back to Central:

- Snapshot resolution: every `org.openrewrite` dependency resolves from CGP at `8.91.0-SNAPSHOT` with Central excluded, including the `org.openrewrite.gradle.tooling` subgroup.
- Release resolution (`-Preleasing`): `latest.release` resolves to `8.90.4` from CGP.
- Release task graph (`final publishPlugins --dry-run` against a throwaway local tag) is unchanged and contains no Sonatype step.
- All `*Sonatype*` tasks are gone; `publishPluginMavenPublicationToCgpRepository` resolves through jar → javadocJar → sourcesJar → POM → publish. The build still configures cleanly with no AWS credentials present.

## Worth flagging

- **Snapshots become credentialed.** `artifacts.codegenomeproject.org` returns 401 anonymously where the Sonatype snapshot repo did not, so anyone consuming `org.openrewrite:plugin:*-SNAPSHOT` without credentials needs a CGP download token. README updated accordingly.
- **The publish step is not exercised by PR CI** — it's gated on `github.event_name != 'pull_request'`, so the first real upload happens on merge to main.
- Now that `cgp` is the only publishing repository, `:plugin:publish` would also push the plugin marker (`publishRewritePluginMarkerMavenPublicationToCgpRepository`), which would make snapshots consumable via `plugins { id("org.openrewrite.rewrite") version "…-SNAPSHOT" }` and retire the `useModule` workaround in the README. Left out as out of scope; happy to add.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ rewrite {
 }
 ```
 
-### Consuming latest snapshots from OSSRH
+### Consuming latest snapshots from the Code Genome Project
 
 To use the latest `-SNAPSHOT` of the `rewrite-gradle-plugin`, update your project's `settings.gradle.kts`:
 
@@ -54,7 +54,11 @@ pluginManagement {
     repositories {
         // ...
         maven {
-            url = uri("https://central.sonatype.com/repository/maven-snapshots/")
+            url = uri("https://artifacts.codegenomeproject.org/maven")
+            credentials {
+                username = "USERNAME"
+                password = "TOKEN"
+            }
         }
         // ...
         // you'll likely also need this if you don't have a pluginManagement section already:
@@ -63,6 +67,10 @@ pluginManagement {
     }
 }
 ```
+
+The Code Genome Project repository requires authentication. Sign in to the Code Genome Project to
+create a download token, then replace `USERNAME` with the email or username you signed in with and
+`TOKEN` with that token.
 
 The plugin can be consumed in your `build.gradle.kts`:
 
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ pluginManagement {
 }
 ```
 
-The Code Genome Project repository requires authentication. Sign in to the Code Genome Project to
-create a download token, then replace `USERNAME` with the email or username you signed in with and
-`TOKEN` with that token.
+The [Code Genome Project](https://codegenomeproject.org/) repository requires authentication.
+[Sign in to get a download token](https://codegenomeproject.org/token), then replace `USERNAME` with
+the email or username you signed in with and `TOKEN` with that token.
 
 The plugin can be consumed in your `build.gradle.kts`:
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -64,6 +68,10 @@ pluginManagement {
 }
 ```
 
+The Code Genome Project repository requires authentication. Sign in to the Code Genome Project to
+create a download token, then replace `USERNAME` with the email or username you signed in with and
+`TOKEN` with that token.
+
```
MBoegers: Maybe add a link to the GCP token claiming process here, or at least to https://codegenomeproject.org/

MBoegers (approved):

PR #472, opened 2026-08-25
Point TestKit test fixtures at the Code Genome Project

### LINKED ISSUE #471: Point TestKit test fixtures at the Code Genome Project instead of Maven Central and Sonatype snapshots
- Follow-up to #467, which cut the main build over to the Code Genome Project. The TestKit fixtures were deliberately left out of that PR because they need credential plumbing and a fork fallback, which is a separate decision.

## What's left

All 35 repository blocks across the test sources are textually identical:

```groovy
repositories {
    mavenLocal()
    mavenCentral()
    maven {
        url = uri("https://central.sonatype.com/repository/maven-snapshots")
    }
}
```

They live in `GradleFixtures.kt`, `RewriteRunTest.kt` (27), `RewriteDryRunTest.kt` (5), `RewriteDiscoverTest.kt` (1), and `RewritePluginTest.kt` (1). There is already a `GradleFixtures.REPOSITORIES` constant, but it is used in exactly one place — the other 34 sites inline the same text.

## Why it matters

These are real Gradle builds launched by `GradleRunner`, so they resolve `org.openrewrite` artifacts over the network at test time. The plugin under test comes from this build via `withPluginClasspath()` (`GradleRunnerTest.kt:45`), but the rewrite runtime libraries do not — `RewritePlugin` creates the `rewrite` configuration and adds `org.openrewrite:rewrite-core:<version>` and friends at runtime, pinned to whatever `generateVersionsProperties` recorded, and the fixture's repositories are what actually serve those jars.

Measured on the last green CI run, of the Maven Central and plugin-portal requests for `org/openrewrite/`:

| source | requests |
| --- | --- |
| main build | 6 |
| TestKit sub-builds | 513 |

- \#467 removed the 6. The 513 are untouched, so the repo still consumes `org.openrewrite` from Maven Central in practice.

## The forward-looking risk

The main build now pins the rewrite version from CGP (currently `8.91.0-SNAPSHOT`), while the fixtures resolve that version from the Sonatype snapshot repo. That works only because `openrewrite/rewrite` still dual-publishes — `8.91.0-SNAPSHOT` was on Sonatype as of 2026-08-25. **When `rewrite` finishes its own cutover and stops publishing snapshots to Sonatype, these tests break**, because the pinned version will not exist in any repository the fixtures know about.

## Suggested approach

1. Thread `codegenomeUsername` / `codegenomePassword` into the test JVMs as system properties, alongside the existing `org.openrewrite.test.gradleVersion` in `plugin/build.gradle.kts`.
2. Make `GradleFixtures.REPOSITORIES` credential-aware: emit the credentialed CGP repository when the properties are present, and fall back to today's Central + Sonatype snapshots when they are not, mirroring the `codegenomeConfigured` pattern the main build now uses. The fallback keeps fork PRs, which receive no secrets, building.
3. Collapse all 35 inline blocks onto the shared constant so this never has to be done site-by-site again.

## Worth knowing while doing this

Three fixtures request recipe artifacts at dynamic versions:

- `rewrite("org.openrewrite.recipe:rewrite-testing-frameworks:latest.release")`
- `rewrite("org.openrewrite.recipe:rewrite-static-analysis:latest.release")`
- `rewrite("org.openrewrite.recipe:rewrite-all:latest.integration")`

Dynamic versions are what make the traffic so heavy: Gradle fetches `maven-metadata.xml`, then walks candidates descending, pulling each POM and each POM's parent `rewrite-bom`. The CI log shows the crawl going `3.44.0` → `3.45.0-SNAPSHOT` → `3.44.0-SNAPSHOT` → `3.43.0-SNAPSHOT`, dragging in `rewrite-bom` `8.89.0`, `8.91.0-SNAPSHOT`, `8.90.0-SNAPSHOT`, `8.88.0` along the way. Combined with `cacheDynamicVersionsFor(0, SECONDS)` and `cacheChangingModulesFor(0, SECONDS)`, nothing is cached between resolutions. Pinning these to concrete versions would cut the request volume sharply and is worth considering as part of the same change.

Note also that `mavenLocal()` is first in the fixture list, which is the escape hatch for testing against a locally published `rewrite` build. Whatever replaces these blocks should keep it.

timtebeek:
Fixes #471.

The main build already resolves `org.openrewrite` from CGP, but the builds TestKit launches resolved the versions it pinned from Maven Central plus the Sonatype snapshots repository. That works only while `rewrite` dual-publishes; once it stops publishing snapshots to Sonatype, the pinned version exists in no repository the fixtures know about.

## What changed

- **Credentials reach the test JVMs.** `plugin/build.gradle.kts` passes `codegenomeUsername` / `codegenomePassword` to every `Test` task as system properties. It uses a `CommandLineArgumentProvider` with no annotated inputs, so the credentials stay out of the task input fingerprint. CI already exports both as `ORG_GRADLE_PROJECT_*`, so no workflow change was needed.
- **`GradleFixtures.REPOSITORIES` is credential-aware.** With the properties present it emits the credentialed CGP repository; without them it emits today's Maven Central plus Sonatype snapshots, which keeps fork pull requests building. `mavenLocal()` stays first either way, so a locally published `rewrite` build still wins.
- **Where the Gradle under test can content-filter (5.1+), CGP is scoped to `org[.]openrewrite.*` and goes ahead of Central.** Ordering it after Central left 234 requests in place as 404s in the first run on this branch. The Gradle 4.10 suite has no filtering API, so there CGP goes last instead; putting an unscoped credentialed repository first sent every third-party lookup through it and ran that suite 37% slower.
- **All 34 inline blocks now come from the fixture**, rendered at the call site's indent so the generated build scripts are byte-identical to what they were. The rendering lives in `TestKitRepositories` next to `GradleProjectSpec`, which lets the `settings.gradle` that `GradleProjectSpec` generates use it too. That was the last remaining Sonatype reference in the test tree.

## What the green run shows

Gradle's own resolution in the sub-builds now goes to CGP: 458 requests to `artifacts.codegenomeproject.org`, and the 11 remaining Sonatype snapshot requests all come from the main build's own `org.openrewrite.tools:jgit` and `rewrite-polyglot` dependencies, not from the fixtures.

Maven Central still sees 238 `org/openrewrite/` requests, down from 443, and they are no longer Gradle resolving the fixture repositories:

- 84 come from the Gradle 4.10 suite, where CGP is last, so Central is asked first and misses.
- The rest are the recipe tasks themselves. Inside `:rewriteRun` / `:rewriteDryRun`, OpenRewrite's own pom downloader probes Central for the same snapshots and 404s; it does not go through Gradle's repository content filters. The tasks succeed regardless, and this is untouched by the fixture change.
- The `latest.release` / `latest.integration` crawl also has to list versions on every repository, Central included.

## Verification

Locally with CGP credentials, `RewriteRunTest.gradleDependencyManagement`, which asserts the exact text of the generated `build.gradle`, passes; that confirms the rendered block matches at both indents and that the emitted repository DSL is valid. The rest of the suite hits Maven Central `HTTP 429` from my machine for third-party artifacts (AGP, Lombok, Checkstyle, `rewrite-testing-frameworks`), so those were left to CI, which is green.

Both long CI runs on this branch stalled for 47 and 64 minutes storing a build cache entry, on `community.develocity.cloud` connection timeouts. That is unrelated to this change; the per-suite times are in line with the baseline.

## Left out

The three fixtures that request recipe artifacts at `latest.release` / `latest.integration` still do. Pinning them would cut request volume sharply, as the issue notes, but it also freezes each one against a rewrite runtime that keeps moving, so it seemed worth a separate decision.

### DIFF
--- a/plugin/src/main/kotlin/org/openrewrite/gradle/GradleProjectSpec.kt
+++ b/plugin/src/main/kotlin/org/openrewrite/gradle/GradleProjectSpec.kt
@@ -109,11 +109,7 @@ class GradleProjectSpec(
                     repositories {
                         gradlePluginPortal()
                         google()
-                        mavenLocal()
-                        mavenCentral()
-                        maven {
-                            url = uri("https://central.sonatype.com/repository/maven-snapshots")
-                        }
+                        ${TestKitRepositories.declarations(24)}
                     }
                 }
             """.trimIndent()
@@ -136,11 +132,7 @@ class GradleProjectSpec(
                     repositories {
                         gradlePluginPortal()
                         google()
-                        mavenLocal()
-                        mavenCentral()
-                        maven {
-                            url = uri("https://central.sonatype.com/repository/maven-snapshots")
-                        }
+                        ${TestKitRepositories.declarations(24)}
                         // jcenter is currently only required for AGP 3.*
                         jcenter()
                     }
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -183,6 +183,22 @@ project.rootProject.tasks.getByName("postRelease").dependsOn(project.tasks.getBy
 
 tasks.withType<Test>().configureEach {
     useJUnitPlatform()
+    // Hand the same Code Genome Project credentials to the builds TestKit launches, so they resolve
+    // the org.openrewrite versions this build pinned in versions.properties from where it got them.
+    // An argument provider without annotated inputs keeps the credentials out of the task's input
+    // fingerprint. Absent the properties, as on fork pull requests, the fixtures fall back to
+    // Maven Central plus Sonatype snapshots.
+    val codegenomeUsername = providers.gradleProperty("codegenomeUsername")
+    val codegenomePassword = providers.gradleProperty("codegenomePassword")
+    jvmArgumentProviders.add(CommandLineArgumentProvider {
+        val username = codegenomeUsername.orNull
+        val password = codegenomePassword.orNull
+        if (username.isNullOrEmpty() || password.isNullOrEmpty()) {
+            emptyList()
+        } else {
+            listOf("-DcodegenomeUsername=$username", "-DcodegenomePassword=$password")
+        }
+    })
 }
 
 val gVP = tasks.register("generateVersionsProperties") {
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -199,11 +199,9 @@ project.rootProject.tasks.getByName("postRelease").dependsOn(project.tasks.getBy
 
 tasks.withType<Test>().configureEach {
     useJUnitPlatform()
-    // Hand the same Code Genome Project credentials to the builds TestKit launches, so they resolve
-    // the org.openrewrite versions this build pinned in versions.properties from where it got them.
-    // An argument provider without annotated inputs keeps the credentials out of the task's input
-    // fingerprint. Absent the properties, as on fork pull requests, the fixtures fall back to
-    // Maven Central plus Sonatype snapshots.
+    // Hand the same credentials to the builds TestKit launches, so they resolve org.openrewrite from
+    // where this build did. An argument provider without annotated inputs keeps them out of the task
+    // input fingerprint.
     val codegenomeUsername = providers.gradleProperty("codegenomeUsername")
     val codegenomePassword = providers.gradleProperty("codegenomePassword")
     jvmArgumentProviders.add(CommandLineArgumentProvider {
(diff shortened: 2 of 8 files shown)


plugin/src/main/kotlin/org/openrewrite/gradle/GradleProjectSpec.kt
```diff
@@ -136,11 +132,7 @@ class GradleProjectSpec(
                     repositories {
                         gradlePluginPortal()
                         google()
-                        mavenLocal()
-                        mavenCentral()
-                        maven {
-                            url = uri("https://central.sonatype.com/repository/maven-snapshots")
-                        }
+                        ${TestKitRepositories.declarations(24)}
```
MBoegers, line 135: I don't like the magic number here. But its is only build tool logic that is invisible to normal users.

MBoegers (approved):

plugin/src/main/kotlin/org/openrewrite/gradle/GradleProjectSpec.kt
timtebeek: I'll go a step further and say I don't like how Gradle does interpolation in text blocks, which is why we need this. :)

